PR #1, opened 2017-09-28
Added more configuration options for ATC - Vault integration

ykrevnyi:


vito:
Thanks a bunch for this! Sorry it took so long to review. I was hesitating as I wasn't sure whether we wanted to go down the route of documenting all the flags to the binary, since they already show up under `--help`, and the way you configure it varies for each distribution. But I think this is still better than nothing.

I may polish things up a bit later and try to find a nice pattern for the more general problem.

PR #6, opened 2017-10-30
add note that DB upgrade must not be combined with stemcell upgrade

philippthun:
otherwise DB upgrade will fail because both old and new postgres packages must be available in order to do a major version upgrade.


vito:
true, thanks!

PR #7, opened 2017-11-02
Fix name of repository that contains the dos.

headcr4sh:
The link had already been updated, but the repo name hat not been updated.


vito:
thanks!

PR #8, opened 2017-11-02
Add link to SonarQube resource

headcr4sh:
This resource can be used to trigger a SonarQube static code analysis (via "sonar-runner") and allows to implement Code quality gates into CI pipelines.


vito:
thanks!

PR #9, opened 2017-11-08
Fixed wrong property for db role in sample manifest

vchrisb:
The role name property for the Postgres database needs to be `name`
[postgres-release customization](https://github.com/cloudfoundry/postgres-release#customizing)


wbean1:
+1, came here to fix the same thing

vito:
sorry about that, thanks!

PR #11, opened 2017-11-13
Changed database property from role to name

z4ce:
`role` doesn't match the spec file of either the postgres release or the concourse release


vito:
fixed by #9, thanks though!

PR #12, opened 2017-11-15
Suggest using PostgreSQL 9.6 for a new installation

beanieboi:
Hey,

ATC uses 9.6 since this commit (https://github.com/concourse/atc/commit/a9453f64de38737006a18495bb1ad64a19e6326c). This updates the docs to use PostgreSQL 9.6 for new installations.

Cheers
Ben

### DIFF
--- a/lit/setting-up/installing.lit
+++ b/lit/setting-up/installing.lit
@@ -313,7 +313,7 @@ a tool that meshes well with Concourse's ideals, buckle up and head over to
 
   services:
     concourse-db:
-      image: postgres:9.5
+      image: postgres:9.6
       environment:
         POSTGRES_DB: concourse
         POSTGRES_USER: concourse


jtarchie:
I think we want the documentation to reflect the minimum version required, which is 9.5. We need to update the docs to reflect that 9.5+ requirement not just in the release notes.

beanieboi:
hey @jtarchie 

I understand your thinking behind that. but that is not how I perceived the installing guides. I understood that 9.5 is the maximum version that is supported by concourse. By accident I saw a commit on ATC that mentions the support of PostgreSQL 9.6. Which confused me a little bit. 

I also went through the docs and they mention that 9.3+ is supported. (https://github.com/concourse/docs/blob/master/lit/setting-up/installing.lit#L99)

Is that still true? Or is 9.5+ now required?

In my personal opinion the installing guides should show the latest possible version. My assumption is that if I'm reading the installing guides I probably want to start a new project and that should go with the latest version possible. I think it's super valuable to mention the 9.5+ support in the docs just in case I'm not able to deploy the latest version of Postgres. 

What do you think?

vito:
I think it's fine to merge this. Maybe we should even have it say just `9`. It said 9.5 even when we supported 9.3, so it's not like we're showing the minimum supported version (we even run 9.6 in prod I think).

9.5+ is required, the docs you linked were out of date (we've fixed it now!)

PR #15, opened 2017-12-01
Call out breaking job spec changes directly in release notes

xtreme-gavin-enns:
It's not really clear from reading the release note entry what (or even that) manifest changes are necessary. The explanation of why the change was made is cute but not critical, and should not detract from making critical information explicit.

### DIFF
--- a/lit/release-notes/v3.7.0.lit
+++ b/lit/release-notes/v3.7.0.lit
@@ -26,6 +26,21 @@
   }
 
   \note{feature,breaking}{
+    Some credentials must be provided that were previously generated.
+    
+    Changes to the bosh job specs are as follows:
+    - atc:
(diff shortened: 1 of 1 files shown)


vito:
I updated the wording, emboldened it, and improved the manifest diff link - is that ok? https://concourse.ci/downloads.html#v370

xtreme-gavin-enns:
Yeah, that's better, honestly I just had a brain-fart and didn't notice that the pul request number was a link. 

I still feel like the release notes should be as explicit and brief as possible though. It's so common to have difficulties with updates, and that would help mitigate it.

Like, there's no real value to the three paragraphs of explanation of the history of why it was implemented that way in the first place, it just obscures the need-to-know information.

PR #17, opened 2017-12-07
add Spinnaker resource

burdzwastaken:


vito:
@jama-pivotal I don't think that merge did what you wanted it to do. :P

PR #26, opened 2018-02-28
setting up: use same Postgres database name everywhere

marco-m:
This PR attempts to reduce confusion in naming the Postgres database as follows:

At line 100, we say

    This assumes you have a local Postgres server running on the default port
    5432 with an `atc` database, accessible by the current user. If your database
    lives elsewhere, just specify the `--postgres-data-source` flag, which is also
    demonstrated below.

If we look at the "below" part (the subject of this PR), the database is called "concourse",
while above the default value is "atc". This PR always uses the "atc" name.


vito:
good catch, thanks!

PR #29, opened 2018-03-07
add k8s concourse custom resource support

srinivasa-vasu:


jama-pivotal:
@srinivasa-vasu  Sorry for the delay. I've manually merged this change into my branch and you should be seeing this change live shortly! thanks for the PR and the new resource!

PR #31, opened 2018-03-15
Add artifactory resource

mbhave:


jama-pivotal:
Manually merging this locally. Thanks for the PR @mbhave and the new resource. You should see this live shortly. Sorry for the delay!

PR #45, opened 2018-04-11
Add the vault-auth-backend-max-ttl entry

baptiste-bonnaudet:
Document https://github.com/concourse/atc/pull/256 new vault-auth-backend-max-ttl flag/environment.


vito:
thanks!

PR #46, opened 2018-04-12
Update cf-cli-resource repo location

patrickcrocker:
I'm no longer maintaining the `Pivotal-Field-Engineering/cf-cli-resource` fork and need to update the docs to point to the original `nulldriver/cf-cli-resource` repo.


jama-pivotal:
done!

PR #48, opened 2018-04-15
Add community PhraseApp resource

tenjaa:
A resource that triggers on changes in a phraseapp project without getting all translations everytime and running into the rate limit


jama-pivotal:
Thanks!

PR #53, opened 2018-04-19
corrected typos.

kukgini:
little change.

### DIFF
--- a/lit/architecture/architecture.lit
+++ b/lit/architecture/architecture.lit
@@ -4,7 +4,7 @@
 
 Concourse is a fairly simple distributed system built up from the following
 components. You'll see them referenced here and there throughout the
-documentation, so you may want to skim this page just to get an idea of what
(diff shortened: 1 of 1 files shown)


lit/architecture/architecture.lit
```diff
@@ -4,7 +4,7 @@
 
 Concourse is a fairly simple distributed system built up from the following
 components. You'll see them referenced here and there throughout the
-documentation, so you may want to skim this page just to get an idea of what
+documentation, so you may want to skip this page just to get an idea of what
```
charlieoleary: The original (skim) is actually correct. Skip basically means to ignore / bypass where as skim means to read quickly. If the meaning of skim isn't clear, it could be changed to `...may want to quickly review...` instead.

kukgini: Oh, Now I understood the intend. 
It's come from my poor language skill. thank you for explain it.

PR #55, opened 2018-04-23
Removed Orange Highlight from the Word 'Tutorial' in Header

sampeinado:
now it will look like this:
screen shot 2018-04-23 at 4 17 56 pm


jama-pivotal:
Thanks!

PR #59, opened 2018-04-26
Add distro packaging resources from Seveas

seveas:
I like to package my software as debs and rpms, so when I switched all my CI
needs to concourse, I had to write a few resources for distributing those
packages. I think they're useful enough for others too.


jama-pivotal:
Thanks!

PR #63, opened 2018-05-03
Add ability to specify concourse version in docker-compose config

mvargeson:
This is probably not super useful for most people but I figured I would throw this upstream just in case. 

The reason I added this was because I wanted to match up with an older version of Concourse that we have deployed so that I don't constantly need to `fly sync` when moving between the two environments. Feel free to close out if it doesn't seem like the right place.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     - PGDATA=/database
 
   concourse:
-    image: concourse/concourse
+    image: concourse/concourse:${CONCOURSE_VERSION:-latest}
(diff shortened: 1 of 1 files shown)


vito:
This is just for the quickstart bit on the website - is that what you're using day to day on your machines? You may want to PR to https://github.com/concourse/concourse-docker/blob/master/docker-compose.yml instead.

mvargeson:
Thanks! I suspected there was something like that but honestly didn't take much of a look. I'll close this.

PR #66, opened 2018-05-16
Add Google Hangouts Resource to community resources

a114m:


jama-pivotal:
Interesting, the dream of google hangouts never died! Thanks for the PR

PR #67, opened 2018-05-22
peckeltw correct link to Amazon Secrets Manager

peckeltw:


jama-pivotal:
Good catch, thanks for the fix @peckeltw !

peckeltw:
You're welcome @jama-pivotal

PR #71, opened 2018-05-25
Update community-resources.lit

rowanjacobs:
Add [bbl-state-resource](https://github.com/cloudfoundry/bbl-state-resource) to the list of community resources.


vito:
thanks!

PR #78, opened 2018-06-17
Add Slack alert resource

arbourd:


jama-pivotal:
Looks like an interesting resource, thanks!

PR #81, opened 2018-06-22
Auth changes

pivotal-jwinters:
These aren't really done, but want to get some early feedback @vito @jama-pivotal

### DIFF
--- a/lit/setup-and-operations/install.lit
+++ b/lit/setup-and-operations/install.lit
@@ -42,17 +42,17 @@ page.
   you might want to just kick the tires a bit and run everything at once. This
   can be achieved with the \code{quickstart} command:
 
-  \codeblock{bash}{{
-  concourse quickstart \\
-    --basic-auth-username myuser \\
-    --basic-auth-password mypass \\
-    --external-url http://my-ci.example.com \\
+  \codeblock{bash}{{{
+  concourse quickstart \
+    --add-local-user myuser:mypass \
+    --main-team-local-user myuser \
+    --external-url http://my-ci.example.com \
     --worker-work-dir /opt/concourse/worker
-  }}
+  }}}
 
   This command is shorthand for running a single \reference{web-node} and
   \reference{worker-node} on the same machine, auto-wired to trust each other.
-  We've also configured some silly basic auth for the \reference{main-team} -
+  We've also configured some local auth for the \reference{main-team} -
   you may want to change that (see \reference{authentication}).
 
   So far we've assumed that you have a local PostgreSQL server running on the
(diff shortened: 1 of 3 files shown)


lit/setup-and-operations/install.lit
```diff
@@ -42,17 +42,17 @@ page.
...
   can be achieved with the \code{quickstart} command:
 
-  \codeblock{bash}{{
-  concourse quickstart \\
-    --basic-auth-username myuser \\
-    --basic-auth-password mypass \\
-    --external-url http://my-ci.example.com \\
+  \codeblock{bash}{{{
+  concourse quickstart \
+    --add-local-user myuser:mypass \
```
vito: The name of this flag throws me off a bit - is it doing a mutation (literally adding a user somewhere) or just configuring the user setup (which I see is global, and then you grant the user access to the main team separately)? If I stop passing the flag, is the user still there?

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
 }
 
+
+\section{
+  \title{Configuring Auth Providers}{auth-config}
+
+  At startup, you will need to configure any local users or third-party auth
+  providers that you wish to use with Concourse.
+
+  In contrast to previous releases of Concourse, \code{users} are now central
```
vito: This type of wording may be better off in the release notes or a blog post - first-time users reading the docs don't really need to know how it's different. The rest of the para is good info to know, I'd just state it outright and we'll cover the "how it's different" in places people will look when upgrading.

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+    \title{Local Auth}{local-auth-config}
+
+    In order to add new users to the system, you can supply the
+    \code{--add-local-user} flag to the \code{concourse web} subcommand like
+    so:
+
+    \codeblock{bash}{{{
+    concourse web \
+      --add-local-user myuser:mypass
+    }}}
```
vito: A paragraph here saying what this effectively does (or doesn't do) may be useful.

For example, "all this does is whitelist the username/password for logging in. Local users don't intrinsically have any capabilities - they must be granted access to individual teams before they can do anything". And then an example of granting access with `set-team` and `--main-team-local-user`. (Or link to the relevant section - I see this is covered already in the 'teams' section.)

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+    infrastructure.
+
+    \section{
+      \title{Creating a GitHub application}
+
+      First you need to \link{create an OAuth application on
+      GitHub}{https://github.com/settings/applications/new}.
+
+      The callback URL will be the external URL of your Concourse server with
+      \code{/sky/issuer/callback} appended. For example, Concourse's own CI
```
vito: Is this right? Ours is `https://ci.concourse-ci.org/sky/dex/callback/github`. (I think the `/github` is redundant, but ours says `dex` not `issuer`.)

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+      \list{
+        \code{--github-host=github.example.com}
+      }{
+        \code{--github-ca-cert=/tmp/some-cert}
+      }
+
+      The GitHub Enterprise host should not contain a scheme, or a trailing
+      slash.
+
+      \codeblock{bash}{{{
```
vito: This codeblock is kind of without context. It can probably just be omitted if we change the above to specify that the flags go to `web`.

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+      \code{/sky/issuer/callback} appended. For example, Concourse's own CI
+      server's callback URL would be
+      \code{https://ci.concourse-ci.org/sky/issuer/callback}.
+    }
+
+    \section{
+      \title{Configuring the client}
+
+      You will be given a Client ID and a Client Secret for your new
+      application. These will then be passed to the following flags:
```
vito: "... following flags to `concourse web`" would be more clear

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+          refresh-token-validity: 3600
+          redirect-uri: https://concourse.example.com/sky/issuer/callback
+      }}
+    }
+
+    \section{
+      \title{Configuring the client}
+
+      You will be given a Client ID and a Client Secret for your new
+      application. These will then be passed to the following flags:
```
vito: Similar feedback as above (specify the command, omit/provide context for the codeblock below).

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+    }
+
+    \section{
+      \title{Configuring the client}
+
+      Configuring generic OIDC for X company's internal OIDC service may look
+      something like:
+
+      \codeblock{bash}{{{
+      $ concourse web \
```
vito: the `$` here isn't in any of the other codeblocks :P

lit/setup-and-operations/install.lit
```diff
@@ -68,6 +68,232 @@ page.
...
+    }
+
+    \section{
+      \title{Configuring the client}
+
+      Configuring generic oAuth for X company's internal oAuth service may look
+      something like:
+
+      \codeblock{bash}{{{
+      $ concourse web \
```
vito: same as above (wow i'm really picking nits)

lit/setup-and-operations/teams.lit
```diff
@@ -6,13 +6,12 @@ Every pipeline and one-off build belongs to a \italic{team}.
 
 A team is simply a conceptual owner and a separate namespace for data in
 Concourse (pipelines, builds, etc.) with its own auth configuration. A team may
-map to GitHub oAuth, basic auth, UAA auth, or whatever other auth methods we
-add in the future.
+whitelist users or groups either from local auth or from an external provider
```
vito: I'd avoid the term "groups" here (since we're abstracting that aspect away with our own CLI) - how about:

A team is simply a conceptual owner and a separate namespace for data in Concourse (pipelines, builds, etc.). Individual user access to a team is determined by how it's configured with `set-team` - either by whitelisting local users or configuring external providers that are enabled for the Concourse server (see \reference{auth-config}).

lit/setup-and-operations/teams.lit
```diff
@@ -45,12 +44,11 @@ as).
     with \code{fly}}, you can run \reference{fly-set-team} to create or
     update other teams.
 
-    To create a new team or update an existing team's auth run:
+    To create a new team or update an existing team's authorized users and groups run:
```
vito: I think "auth" was fine before (avoid 'groups'). Maybe "auth settings"?

lit/setup-and-operations/teams.lit
```diff
@@ -140,271 +138,146 @@ as).
   will be given a choice when logging in as to which one they would like to
   use.
 
+  \warn{
+    In order to use any of the auth providers listed below, you must first
+    enable the provider by configuring any required information at deployment
+    time as per the \reference{install} guide.
```
vito: Could deep-reference `auth-config` here.

lit/setup-and-operations/install.lit
pivotal-jwinters: We're trying to remove all the outward references to `dex`. Figured `issuer` was a nice generic name.

PR #84, opened 2018-06-28
Add http resource.

aequitas:


jama-pivotal:
Interesting! Thanks for writing the resourec

PR #86, opened 2018-07-05
updated documentation to have example using bosh for deployment.   Ad…

fewknow:
Added example for bosh deployment manifest using vault cred management in the atc job.  Also linked to the documentation for this.  Updated wording to periodic vault token as this was the type that I had to use to connect vault backend to my concourse.


vito:
I recognize that this leaves a bit of a hole, but we intentionally don't intermingle BOSH docs with the binaries. We kept getting feedback that new users would see BOSH and run away afraid of having to learn a new tool (even though we made it super clear that you don't have to).

Thanks though!

PR #87, opened 2018-07-06
the emerald-squad is replacing cf-blue-green-resource with cf-zero-do…

ukabu:
…wntime-resource

We've deprecated the [cf-blue-green-resource](https://github.com/emerald-squad/cf-blue-green-resource) and replacing it with an alternative that has more features [cf-zero-downtime-resource](https://github.com/emerald-squad/cf-zero-downtime-resource).


vito:
thanks!

PR #99, opened 2018-08-17
Update vault docs to include example to set env var in task

walterscarborough:
Signed-off-by: Sarah McAlear <smcalear@pivotal.io>

### DIFF
--- a/lit/setup-and-operations/operation/creds.lit
+++ b/lit/setup-and-operations/operation/creds.lit
@@ -53,6 +53,23 @@ values are not present, the action will error.
   \omit-children-from-table-of-contents
 
   \section{
+    \title{Usage With Tasks}
+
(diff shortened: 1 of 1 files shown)


walterscarborough:
We thought that it was very hard to set up vault with a task, and that adding this example would help clarify that.

Thanks,

Walter and Sarah

vito:
Hi! Thanks for taking the time to submit this, and sorry I never got around to giving feedback on this or merging it. 

Recently I've rewritten the Vault documentation to be more...well, useful. I just realized one thing still missing is showing an example of how to set and use credentials, so I opened https://github.com/concourse/docs/issues/174 and then remembered this PR. Since I can't merge it anymore and you've probably since moved on with your life, I figure I'll just close this and start on the #174 soon. Hope that's OK.

Thanks again!

PR #112, opened 2018-09-20
Allow the quickstart compose file's envvars to be overridden

jpluscplusm:
The quickstart `docker-compose.yml` file hardcodes some parameters. Docker Compose supports an interpolation mechanism which allows the same defaulting to happen, but in a way which permits the user to modify these parameters when running docker-compose.

Making this change will allow this Compose file to be used by more consumers, whilst still enabling the same low-friction quickstart workflow.

Alternative mechanisms, such as adding a `docker-compose.env` file and wget'ing it as part of the quickstart, are possible - and don't increase the quickstart friction very much. This commit implements the mechanism which doesn't change the quickstart experience *at all*.

This PR also makes it clearer to the prospective overriding user how to set the user password they want.

Tests
-----

I'm not including any tests. Here's my rationale:

- if there already exist any CI/etc tests of the `docker-compose.yml` file, then this PR will either break them, or it won't. If it doesn't, and the status quo experience (i.e. no externally set envvars) doesn't change, then by definition the status quo has been maintained. Even if the additional functionality intended to be implemented here doesn't work, nothing of value was lost.

- if there aren't any tests of this file, then implementing a test framework to run up machines on which to run this set of containers seems like something that needs project-level organisation, and isn't in scope for a drive-by PR'ing. IMHO :-)


vito:
Hmm I'm a bit lukewarm on this. Thanks for submitting it, obviously, and sorry for the delay - we're in flux right now with https://github.com/concourse/concourse/issues/2534 and that's slowing down our PR influx.

The added flexibility is nice but I'm not sure we want the quick start being used for much more than the initial tire-kicking, and I'm afraid that by opening it up to more configuration we'll make it look more like a real-world Concourse deployment template. We don't intend to maintain this file for use by downstream deployments - it's just there as an example.

More real-world use cases should probably use a full-fledged Docker Compose file (and forego `quickstart` if you don't want your user tokens expiring on every restart), such as the one found in https://github.com/concourse/concourse-docker. Which also has things hard-coded, but that one would make a lot more sense to become flexible. Maybe a PR there would make more sense?

Thanks again and sorry for the wait!

jama-pivotal:
Closing for now based off @vito 's feedback. Thanks for making hte PR though!

PR #116, opened 2018-09-28
Update teams.lit to include details for teams in fly cli -d. Found it h…

doughoke:
Added documentation for fly teams -d and notes with debugging team auth configurations.

### DIFF
--- a/lit/setup-and-operations/teams.lit
+++ b/lit/setup-and-operations/teams.lit
@@ -93,6 +93,18 @@ against the teams that have granted them access.
   }
 
   \section{
+    \title{\code{fly teams -d}\aux{: With Details}}{fly teams -d}
+
+    To list all the teams with authentication details and members, run:
+
+    \codeblock{bash}{{
+    $ fly -t example teams -d
+    }}
+
+    This can be helpful when debugging OAuth, OIDC groups or listing all individual members.
+  }
+
+  \section{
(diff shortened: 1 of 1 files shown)


jama-pivotal:
Thanks for submitting this @doughoke , and really sorry for the delay in responding to you. 

Wanted to preface my feedback with: We really should do a better job at documenting some of the `fly` CLI flags and also `concourse` binary flags. Sure, people can use `--help` but it's nice if some of this stuff was searchable and google-able.

Anyways, that said, can I make a suggestion to de-emphasize the `fly teams -d: With Details` as a section and just use a `\bold{}` instead. I think it makes a bit more sense in the flow of this documentation to not make this a "section" in our docs. Makes the index looks off since really its more a commentary / tip on `fly teams`

screen shot 2018-10-22 at 10 47 41 am

lit/setup-and-operations/teams.lit
```diff
@@ -92,6 +92,18 @@ against the teams that have granted them access.
     This can be useful if you've forgotten your team name.
   }
 
+  \section{
```
jama-pivotal: Suggest changing from `\section` to just a `bold`

doughoke: OK - I can make that change. I will roll it up today.

jama-pivotal:
Thanks!

PR #123, opened 2018-11-02
added packer resource

eedwards-sk:
Adds the snapkitchen packer resource as per https://discuss.concourse-ci.org/t/concourse-packer-resource/777

I noticed the list could probably use to be alphabetized, but I just placed mine at the bottom to be safe.


jama-pivotal:
Thanks!

PR #124, opened 2018-11-04
Explain how to work around golang crypto pickiness for RSA keys

marco-m:
See for details:

* https://github.com/concourse/concourse/issues/2592
* https://discuss.concourse-ci.org/t/installation-issues/675/


marco-m:
Fixes https://github.com/concourse/concourse/issues/2592
@drahnr opinions ?

vito:
Good catch! Is there any way we can fix our usage of `ssh-keygen`, or perhaps just use another tool that doesn't have this problem instead? Should we just generate them with `openssl`? I'm a bit worried putting troubleshooting in the installation docs since I'd prefer to keep the introductory material "happy path" for new users, and leave troubleshooting to us to either fix or document somewhere easy to find (which admittedly doesn't exist at the moment, besides GitHub issues).

drahnr:
@vito in general, you are right. In this case, I think it is worth mentioning. The intro guide is pointing with a shiny arrow down a cliff, and for most people who are just getting started, this has potential to steer them away from concourse.

@marco-m  for the tutorial, I think it'd be better to keep it short and just show the correct line for `ssh-keygen` with its working arguments and add a oneliner warning containing the error with a very brief explanation, just as reference and so duckduckgo, google and friends index it.

marco-m:
From my experiments there is no way to fix the usage of `ssh-keygen` unfortunately.
I understand the reasoning about not wanting to put troubleshooting in the installation guide, I was dubious too.

We could just replace the usage of `ssh-keygen` with the usage of openssl, as @vito mentions.

I will propose version 2.

vito:
@drahnr I'm not proposing that we do nothing, I'm proposing that we just fix the instructions. :slightly_smiling_face:

@marco-m Sounds good!

If people still run into this after making the instructions more foolproof, perhaps we could cite the Go issue in an `\aside{}` or something - I'm OK with adding tips and pointers, I just don't want workarounds to dominate the "golden onramp" over time, content-wise.

vito:
Alternatively there could be low-hanging fruit here UX-wise - perhaps a `concourse generate-keys` command that uses Go libraries to generate the keys, ensuring Go will be happy with them? That would also be more portable to Windows. Hmm....

drahnr:
@vito I didn't grasp that, sorrey :]

The proposal regarding a `concourse generate-keys` sounds pretty appealing to me :)

vito:
@drahnr No worries!

I made an issue for `concourse generate-key`, I haven't prioritized it anywhere but I gave it a 'good first issue' label: https://github.com/concourse/concourse/issues/2779

marco-m:
@vito rebased and pushed new commit. I left old commit for comparison only, this PR should be merge-squashed.

I also found a gotcha: `openssl genpkey` doesn't change the file mode, so `ssh-keygen` correctly refuses to run. I added also a `chmod`.

I unrolled the shell loop on purpose since you mentioned Windows; now the code is slightly more portable although more verbose.

@drahnr Since this is going in the getting started guide, could you help me with the double-check ?
I would appreciate if you could emulate a new user: copy and paste the shell commands and verify that the generated keys work with Concourse ? I already tested but a double-check is worth I think :-)

marco-m:
Mhh I still get some errors, I am confused. Let me check.

drahnr:
@marco-m I'll give it a whirl this week, hopefully tomorrow

eedwards-sk:
I lost an entire day to this because the error is uncaught and the stack trace is extremely cryptic (no pun intended): https://github.com/concourse/concourse/issues/2526

+1 to this...

vito:
the docs now use `concourse generate-key` everywhere :raised_hands:

PR #127, opened 2018-11-10
Adjusts for Go modules

osis:


vito:
thanks! ended up changing things around myself to clear out all the submodules, too

PR #128, opened 2018-11-13
Correct link to `CONTRIBUTING.md`

jpluscplusm:
This fixes a 404 I encountered whilst using the Concourse website.

(I *believe* the link now goes to the intended document!)


jama-pivotal:
Good catch!! Thanks for making the fix

PR #134, opened 2018-12-11
update index.lit

iamjarvo:
Surround full word in code block

### DIFF
--- a/lit/index.lit
+++ b/lit/index.lit
@@ -147,7 +147,7 @@
   fetch bits from, and push bits to.
 
   For example, a resource with type \code{git} refers to a git repository,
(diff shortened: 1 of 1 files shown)


jama-pivotal:
I believe the styling is intentional since `get` and `push` are the Concourse-specific yml verbs. @vito could correct me if I'm wrong though.

iamjarvo:
I'd vote for removing them but fine with whatever the decision is. I figured there might have been a reason for the partial `highlight`. It looked funky though

jama22:
Just gonna make the call on this one and leaving it as is. Thanks for the feedback !

PR #137, opened 2018-12-30
Add Ofcourse Resource Generator

cloudboss:
This is more of a meta-resource than a useful resource on its own. Using this tool and library, you can create a "noop" resource that can be extended into something useful.


jama-pivotal:
@cloudboss interesting! You're right thought it doesn't quite belong on this page. And it doesn't quite belong on the Tutorials page either (https://concourse-ci.org/tutorials.html) 

@vito how would you feel if I renamed the Tutorials page to `Tutorials & Tools` and we add a reference to Ofcourse there instead?

vito:
@jama-pivotal Sounds good to me!

jama-pivotal:
Merging for now and will conduct the suggested renaming in a separate issue

PR #138, opened 2019-01-02
Fix a link for metrics

cappyzawa:
https://github.com/concourse/atc is no longer maintained because it has been archived.
I fix this link.


jama-pivotal:
Thanks!

PR #140, opened 2019-01-17
document quickstart's need for sudo

mdhender:
Adds a note to the Quickstart explaining why we might need sudo.

Resolves https://github.com/concourse/concourse/issues/3065


vito:
perfect, thanks!

PR #146, opened 2019-01-22
external task variables rework in 5.0

ralekseenkov:


ralekseenkov:
feel free to edit/reword before merging.
I've tried to find all places related to vars and params in the docs, and ensure they are consistent with the code changes.

lit/setup-and-operations/operation/creds.lit
```diff
@@ -26,20 +26,12 @@ values are not present, the action will error.
   The following configurations can be parameterized with a credential manager:
 
   \list{
-    \reference{resource-source} under \reference{resources} in a pipeline
```
vito: I think I'll restore this to how it was and just add `file`/`config` - these are structured as a dry list like "`field-name` in (location)" which I find a bit clearer for knowing where exactly I can use credential management.

For example, you can still use credential management on `params` on a task step, which is kind of awkward to group under 'tasks' since that now refers to the whole configuration

vito: Does `vars:` also support interpolation via the credential manager, for its own values? I noticed that's not listed here. Looking at the code, it actually doesn't look like it.

e.g.:

```
task: foo
file: my/task.yml
vars:
  foo_bar: ((baz))
```

If I were to configure a pipeline with the above task, would it be able to pull `((baz))` from the credential manager?

Granted, the task config itself could just use the var directly and skip `vars:`, but this could still be useful for when the task needs it under a different name (imagining a scenario where the task's var is a generic name that is satisfied by a more specific var name decided at the pipeline level).

ralekseenkov: Makes sense. If vars are not currently interpolated, we should probably add a line of code to interpolate them before releasing 5.0.

vito: It might actually work as-is; I'll try it out before merging.

vito: Verified that this works - the template package does indeed seem to recursively interpolate, so it works just fine.

One gotcha: the current implementation seems to scope the credential manager vars over task-step-provided `vars:`, meaning you can't override them. This may be something to fix in the future. Here's how I tested it (using a dummy credential manager I whipped up):

https://github.com/vito/concourse/commit/3ddf159a41e7255760733a1b546a3cd725033db1

It seems to be down to the ordering defined here:

https://github.com/concourse/concourse/blob/0b9f266262d231a499ae3e03e8bc68ab404a4dc3/atc/exec/garden_factory.go#L148

I tried swapping it and the results were actually also surprising, I think because `vars:` isn't interpolated first - it printed this:

```
+ echo some_var is hi_im_some_var
some_var is hi_im_some_var
+ echo some_renamed_var is overruled
some_renamed_var is overruled
+ echo pipeline_var is overruled
pipeline_var is overruled
+ echo ubuntu_tag is xenial
ubuntu_tag is xenial
```

If `vars:` were to be evaluated on its own first, I would have expected `some_renamed_var is SOME_VAR`, but instead it ends up seeing its own vars when interpolating and resolves to its locally defined value.

This is a bit of an edge case I think, so it's not a blocker for the release or anything, just wanted to note it somewhere for posterity.

Here's a commit demo'ing that:

https://github.com/vito/concourse/commit/35683a1080e1f37632fd02401d29459270395686

vito:
merged manually and made some amendments on top - thanks!

Ianwww:
To be sure the docs have been updated for 5.0 but the functionality exists in the latest release?

vito:
@Ianwww this functionality is not yet shipped, it will be in 5.0

PR #164, opened 2019-03-01
v5.0 restructure and new docs

vito:
* Move `Download` to the top of the nav. I think this should be more eye-catching; most software sites have it as a primary nav element.
  * I just noticed it's at the top...and I also just noticed I never look at the top and it still has a 'references' link. Should we re-think having two navs?
* Rewrote and consolidated auth docs; rather than documenting providers in 'install' and team configuration elsewhere, there is one section for each provider that then covers both together, one as *authentication* and the other as *authorization*.
* Updated the install instructions for the new distribution format (archive).
* New install sections covering `web` and `worker`, including all information relevant to operations. (I've cannibalized the Topology section into these sections.)
  * Install instructions now use (and document) `CONCOURSE_` env vars rather than flags - this should be easier to copy-paste for Docker folks and makes it easier to document individual configs (rather than showing partial commands).
* New install section covering PostgreSQL basics - this could be expanded for any config tuning we want to add in the future.
* Link to resources like [`foo` resource](#) instead of [Foo resource](#)
* Switch `<title>` to `(title) - Concourse CI` instead of `Concourse CI - (title)` - easier to skim tabs this way.
* Add a section documenting container placement.
* Move 'Global Resources' under 'Operation' and add `(experimental)` to the title.


vito:
I feel bad about that `wip` commit, heh - maybe I'll rebase and split that up.

lit/index.lit
```diff
@@ -13,25 +13,27 @@
   \link{CI}{https://en.wikipedia.org/wiki/Continuous_integration}/\link{CD}{https://en.wikipedia.org/wiki/Continuous_delivery}.
 }
 
-\quick-start{
+\right-side{Quick Start}{
```
jama22: Wondering if we could perhaps make all sidebar code blocks use the black? the grey on grey might be hard to distinguish on lower color-quality monitors.

vito: done

cirocosta:
Hey, I was looking at the TOC under `learning`, and the little misalignment in the `ol li`'s was bugging me a bit 😬  

screen shot 2019-03-06 at 8 42 55 am

*(in the screenshot I added the padding-left there to the first element)*

Is that intentional? 

thx!

vito:
@cirocosta those example sections should probably just use `\omit-children-from-table-of-contents` - more generally I'm planning on cleaning up/reworking the learning and project sections. gonna keep iterating on it post-5.0 and merge what's here for now

PR #183, opened 2019-03-21
Concourse 5 allows overlapping paths

UncleSamSwiss:
If I understood https://concourse-ci.org/download.html#v500-note-16 correctly, we can now have overlapping inputs and outputs.
Thus I suggest to change the wording here (or at least explain that it is now allowed in some cases).
Please feel free to rewrite my proposal if you find better wording.


vito:
good catch - I've removed the whole paragraph in https://github.com/concourse/docs/commit/9053a63d7acde585ae5cd4522b2dcd09e58de4a8 :+1:

PR #185, opened 2019-03-22
use JSON files to describe community resource types

mockersf:
Instead of listing all resource types in the .lit file, move them to json files and generate the list using a plugin.

The json files can contains either one or many resource type descriptions.

This should make it easier to update the list or add features to the list (indexing, search, sort, ...)

It can be viewed [here](https://mockersf.github.io/docs/community.html#community-resources) and should look exactly alike

It's my first time writing go, so any comment is welcome


vito:
Merged, refactored the Go code a little bit - it was pretty much fine as it was, I just moved code around and fixed a few things to be more conventional Go. :+1: Thanks again!

PR #201, opened 2019-04-30
nav rework

vito:
most context is in the commit messages!


lit/examples.lit
```diff
@@ -13,3 +26,8 @@
 
 \split-sections
 \table-of-contents
+
+For a real-world example, check out \link{Concourse's own
```
jama22: Suggested change:

For a real-world example, check out Concourse' own pipeline in our \link{production environment}{https://ci.concourse-ci.org/teams/main/pipelines/concourse} or read it in \link{code}{https://github.com/concourse/pipelines/tree/master/ci.concourse-ci.org}:

vito: i'll add a link to its source (it's actually at https://github.com/concourse/concourse/blob/master/ci/pipelines/concourse.yml) 👍

PR #202, opened 2019-05-01
Fixes the hover state color for header links

osis:
Links were disappearing on Safari. 

concourse-docs-links

giphy (22)


vito:
are you SUUUURE this isn't just a cache issue? i saw this originally but it was fixed upon refresh. if you load up concourse-ci.org and cmd+shift+r does this still happen?

PR #205, opened 2019-05-07
in master, quickstart will not start without specifying worker work-dir

nrvale0:
Quickstart will not start unless worker's work-dir is specified. This is for the code currently in master branch. The released code will not start due to non-specified TLS options.


vito:
I don't think this is necessary, or at least it won't be once we ship v5.2. :thinking: We definitely broke quickstart, but it's a different required flag:

```
concourse_1     | the required flag `--worker-tsa-worker-private-key' was not specified
docs_concourse_1 exited with code 1
```

Pointing the image to the latest `concourse/concourse-rc` seems to work (without specifying the work-dir), so I'm going to close this. Thanks though!

PR #211, opened 2019-05-28
Fix download link disappearing

### LINKED ISSUE #210: Download link disappears in certain resolutions
The download link disappears on the homepage when the width of the browser window is between about this range (apologies for my poor CSS/HTML debugging skills): `1201-1284px`

Happens with the latest versions of Chrome 74 and Firefox 67.

Download link is visible:
Screen Shot 2019-05-27 at 9 00 22 PM
It disappears:
Screen Shot 2019-05-27 at 9 00 33 PM
Link is visible again once on a mobile resolution:
Screen Shot 2019-05-27 at 9 00 42 PM
Does disappear again though (probably not a big deal for a mobile user though):
Screen Shot 2019-05-27 at 9 05 40 PM

taylorsilva:
Fixes #210

Only tested on Chrome 74 and Firefox 67. I was able to build and render the site without any other visible issues.

### DIFF
--- a/css/booklit.css
+++ b/css/booklit.css
@@ -743,6 +743,7 @@ h3 .anchor {
 }
 .top-nav {
   position: relative;
+  white-space: nowrap;
 }
 .search-input {
   z-index: 101;
--- a/less/booklit.less
+++ b/less/booklit.less
@@ -921,6 +921,7 @@ h3 .anchor {
 
 .top-nav {
   position: relative;
+  white-space: nowrap;
 }
 
 .search-input {


css/booklit.css
```diff
@@ -743,6 +743,7 @@ h3 .anchor {
 }
 .top-nav {
   position: relative;
+  white-space: nowrap;
```
vito: This change needs to be made to `less/booklit.less` instead and then compiled :slightly_smiling_face:

taylorsilva: Knew I'd miss something. I don't understand this newer css stuff :(

taylorsilva: I initially misunderstood your comment. Thought I had to _only_ make the change in the `/less` directory, but I had to make the change in both `less` and `css`. I needlessly deleted my commit haha.

taylorsilva:
I initially misunderstood your comment. Thought I had to _only_ make the change in the `/less` directory, but I had to make the change in both `less` and `css`. I needlessly deleted my commit haha.

PR #213, opened 2019-05-30
Add documentation for Kubernetes credential management

### LINKED ISSUE #96: No documentation for Kubernetes credential manager
Just found out that there's a Kubernetes creds plugin, which is completely undocumented. Might help to add a bit about this.

cirocosta:
fixes #96

### DIFF
--- a/lit/docs/operation/creds/kubernetes.lit
+++ b/lit/docs/operation/creds/kubernetes.lit
@@ -1,0 +1,128 @@
+\title{Kubernetes Credential Manager}{kubernetes-credential-manager}
+
+\use-plugin{concourse-docs}
+\omit-children-from-table-of-contents
+
+
+
+Concourse can be configured to pull credentials from \link{Kubernetes
+\code{secret}
+objects}{https://kubernetes.io/docs/concepts/configuration/secret}.
+
+To configure this, either enable the in-cluster client by setting the
+following env on the \reference{web-node}:
+
+\codeblock{bash}{{{
+CONCOURSE_KUBERNETES_IN_CLUSTER=true
+}}}
+
+or set the path to a \code{kubeconfig} file:
+
+\codeblock{bash}{{{
+CONCOURSE_KUBERNETES_CONFIG_PATH=~/.kube/config
+}}}
+
+\section{
+  \title{Credential lookup rules}{kubernetes-credential-lookup-rules}
+
+  When resolving a parameter such as \code{((foo))}, Concourse will resolve in the following order in the namespace configured for that team:
+
+  \list{
+    \codeblock{c}{{{
+    Name:         PIPELINE_NAME.foo
+    Namespace:    concourse-TEAM_NAME
+    Type:         Opaque
+
+    Data
+    ====
+    value:        32 bytes
+    }}}
+  }{
+    \codeblock{c}{{{
+    Name:         foo
+    Namespace:    concourse-TEAM_NAME
+    Type:         Opaque
+
+    Data
+    ====
+    value:        32 bytes
+    }}}
+  }
+
+  You can specify the field to grab via \code{.} syntax, e.g. \code{((foo.bar))}:
+
+  \list{
+    \codeblock{c}{{{
+    Name:         PIPELINE_NAME.foo
+    Namespace:    concourse-TEAM_NAME
+    Type:         Opaque
+
+    Data
+    ====
+    bar:          32 bytes
+    }}}
+  }{
+    \codeblock{c}{{{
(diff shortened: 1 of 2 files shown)


lit/docs/operation/creds/kubernetes.lit
```diff
@@ -0,0 +1,251 @@
...
+
+  As an example, the following sample RBAC Kubernetes could be created to allow
+  \code{web} to interpolate credentials for "team-a" and "team-b":
+
+
+  \codeblock{yaml}{{{
+  ---
+  # Identifies the `web` service as an actor.
+  #
+  # ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/
```
cirocosta: @vito do you think this kind of comment is too much? Providing references to the document from the official Kubernetes website?

lit/docs/operation/creds/kubernetes.lit
```diff
@@ -0,0 +1,251 @@
...
+        serviceAccountName: web
+        containers:
+          - name: web
+            image: "concourse/concourse:5.2.0"
+            args: [ web ]
+            env:
+              - name: CONCOURSE_KUBERNETES_NAMESPACE_PREFIX
+                value: "myprefix-"
+            # ...
+  }}}
```
cirocosta: this YAML file is quite big 😬 I'm not sure if this would be the best way of presenting it - I tried separating them into separate code blocks, but I feel that this way looks a bit better (and it's quite common to see docs like this when it comes to k8s).

wdyt?

lit/docs/operation/creds/kubernetes.lit
```diff
@@ -0,0 +1,251 @@
...
+    }}}
+  }
+
+  The prefix prepended to the namespace used by Concourse to search for secrets
+  (in the examples above, \code{concourse-}) can be changed by configuring the
+  following in the \reference{web-node}:
+
+  \codeblock{bash}{{{
+  CONCOURSE_KUBERNETES_NAMESPACE_PREFIX=some-other-prefix-
+  }}}
```
cirocosta: I was a bit divided by letting this configuration piece here - it seemed quite far from the other two configurations at the top of the page 🤔  any thoughts on this?

lit/docs/operation/creds/kubernetes.lit
cirocosta: Screen Shot 2019-05-30 at 15 15 28-fullpage

lit/docs/operation/creds/kubernetes.lit
```diff
@@ -0,0 +1,251 @@
...
+  If an action is being run in a one-off build, Concourse will not include the
+  pipeline name in the secret that it looks for.
+
+}
+
+\section{
+  \title{Configuring Kubernetes RBAC}
+
+  As the Web nodes need to retrieve secrets from namespaces that are not its
+  own, it needs extra permissions to do so.
```
marco-m: `web nodes` are plural but `its` and `it` are singular.

lit/docs/operation/creds/kubernetes.lit
vito: I think it's fine - it's the same as with the other credential managers. I'm not sure how often folks will set this but since it has a sane default I think it can be "below the fold".

lit/docs/operation/creds/kubernetes.lit
vito: Would an alternative be to split it up and move the comments into the Booklit document? Or is having it all in one YAML document useful for copy-pasting and submitting everything at once?

lit/docs/operation/creds/kubernetes.lit
vito: Nah that's super useful. :+1:

lit/docs/operation/creds/kubernetes.lit
cirocosta: ooooh good point, those side comments would be nice. I'll give them a try here 👍  


> Or is having it all in one YAML document useful for copy-pasting and submitting everything at once?

hmmm I saw a couple of times people having multiple yaml blocks altogether when it comes to k8s, but I'd personally split 😅

cirocosta: after splitting, seems easier to follow; wdyt?

Screen Shot 2019-06-20 at 16 34 44-fullpage

lit/docs/operation/creds/kubernetes.lit
cirocosta: truuuue, thanks!!

PR #227, opened 2019-07-17
document fly completion

pivotal-jamie-klassen:
Since it was added to v5.4.0 by concourse/concourse#4012


pivotal-jamie-klassen:
@xtremerui there actually is help text for `fly completion`, but there's a bug right now, which concourse/concourse#4136 fixes. I didn't want to mention the whole list of supported shells because it would incur maintenance cost for the docs whenever that list changes.

nader-ziada:
lgtm

PR #233, opened 2019-07-24
typo fixed

mohsen0:


lit/docs/operation/creds/aws-secretsmanager.lit
```diff
@@ -142,5 +142,5 @@
 
   The leading \code{/concourse} can be changed by specifying
   \code{--aws-secretsmanager-pipeline-secret-template} or
-  \code{--aws-secretsmanager-team-secret-template} variables.
+  \code{CONCOURSE_AWS_SECRETSMANAGER_TEAM_SECRET_TEMPLATE} variable.
```
vito: what part of this was a typo? :thinking: kinda weird that this changes one flag to the env var equivalent but then leaves its sibling as a flag (`--aws-secretsmanager-pipeline-secret-template`)

deniseyu: I think it should just be in all caps on this line, the characters are not actually incorrect (at least according to `concourse web --help`). Maybe we could say more clearly that the second one is an environment variable though. Further up on this page we provide both the lowercase flag and uppercased env var!

mohsen0: sorry I miss read that

PR #235, opened 2019-08-08
fix a couple of typos

vhf:
Hi there, I noticed a couple of typos :)


vito:
actually, @pivotal-jamie-klassen it looks like the CI build needs creds for the dummy github bot.

alternatively we could have it only fetch the version info of the token is specified. :thinking:

PR #238, opened 2019-08-19
Update PHP example composer-setup.php sha384 value.

joegit1344:
Going through the PHP example, I noticed a failure similar to what was seen in the Concourse CI. This updates the SHA384 value and fixes the issue.


osis:
Awesome! Thanks @joegit1344! 🙇

PR #241, opened 2019-08-28
Clarified that tasks.run.path is relative to working dir.

mikijov:


deniseyu:
while https://github.com/concourse/concourse/pull/4299 is still in flight, it might still be useful to have this clarification in the docs. I'll just add a note to the open PR to make sure that this line is updated when that gets merged. thanks @mikijov for this PR!

PR #247, opened 2019-09-05
Fix reference

aryzing:
In the original description, a non-existent variable `name` was being referenced in the description of the `resource` definition. This PR fixes that description and correctly references `put`.


deniseyu:
hey @aryzing, this change seems reasonable. To make this mergeable, you'll just need to sign off on the commit! You can do something like 1`git commit --amend -s` and then force push the same commit to this branch.

aryzing:
signed off :)

Issue #244, opened 2019-09-04
Document minimum number of connections that Postgres must allow


cirocosta:
Hey,

Despite the fact that one can observe that a certain number of connections are kept by the pools that we create, and that over time ATC goes over that stable amount, we don't document anywhere what's the maximum that it can reach (32 per backend).

Although the default BOSH postgres release gives us some [good number of max connections by default (500)](https://github.com/cloudfoundry/postgres-release/blob/602e0edb668c91a4bfa1cb4a4462bcce27eb055c/jobs/postgres/spec#L62-L64), that's not necessarily true to other distributions (including CloudSQL, which has a variable number of max conns depending on instance type: https://cloud.google.com/sql/docs/postgres/flags#postgres-m).

I can see that this could be a matter of providing a simple calculation that one could perform: #web_replicas * 64 (or whatever ends up being true after https://github.com/concourse/concourse/issues/4025) in the [Running a PostgreSQL section](https://concourse-ci.org/postgresql-node.html).

Wdyt?

thanks!

cirocosta:
(might be worth documenting that those values are configurable: https://github.com/concourse/concourse/blob/c5dc52503a651c3858f756b22946d612b5bfa367/atc/atccmd/command.go#L102)

pivotal-jamie-klassen:
@cirocosta is this fixed by #315?

cirocosta:
yepp!

PR #253, opened 2019-09-17
mention relationship w postgres max connections

pivotal-jamie-klassen:
almost definitely needs wordsmithing, but let's start somewhere!

fixes concourse/docs#244

### DIFF
--- a/lit/docs/install/web.lit
+++ b/lit/docs/install/web.lit
@@ -158,15 +158,6 @@ as performing all pipeline scheduling. It's basically the brain of Concourse.
     To run a cluster of \reference{web-node}s, you'll first need to ensure
     they're all pointing to the same PostgreSQL server.
 
-    Furthermore, you may wish to configure the max number of parallel database
-    connections that each node makes. The number you choose should not be
-    greater than the maximum number of simultaneous connections your Postgres
-    server will allow. For example:
-
-    \codeblock{bash}{{{
-    CONCOURSE_MAX_OPEN_CONNECTIONS=32
-    }}}
-
     Next, you'll need to configure a \italic{peer address}. This is a DNS or IP
     address that can be used to reach this \code{web} node from other
     \code{web} nodes. Typically this uses a private IP, like so:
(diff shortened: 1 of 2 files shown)


lit/docs/install/web.lit
```diff
@@ -190,6 +181,19 @@ as performing all pipeline scheduling. It's basically the brain of Concourse.
     \code{web} nodes in the cluster to ensure consistent results.
   }
 
+  \section{
+    \title{Connection pools}{web-connection-pooling}
+
+    Furthermore, you may wish to configure the max number of parallel database
+    connections that each node makes. The number you choose should not be
```
cirocosta: ooh I don't think this gives the precise meaning of `CONCOURSE_MAX_OPEN_CONNECTIONS` - reading this statement, I'd assume that setting that to 10 would force a `web` node to not go past 10 conns to `db`, but that's not true - the actual max would be `10 * $number_of_pools`

```go
	MaxOpenConnections int `long:"max-conns" description:"The maximum number of open connections for a connection pool." default:"32"`

// ...

	apiConn, err := cmd.constructDBConn(retryingDriverName, logger, cmd.MaxOpenConnections, "api", lockFactory)
	if err != nil {
		return nil, err
	}

	backendConn, err := cmd.constructDBConn(retryingDriverName, logger, cmd.MaxOpenConnections, "backend", lockFactory)
	if err != nil {
		return nil, err
	}

```

I wonder if we should also document that each web node has two pools 🤔 

wdyt?

thx!

matthewpereira: Should we rename this property entirely (to something like `CONCOURSE_MAX_OPEN_CONNECTIONS_PER_POOL`) or change its behavior to be more in line with the name of the property?

Or is it sufficient to amend our instructions to read:
```
Furthermore, you may wish to configure the max number of parallel database connections per pool that each node makes.
```
...so that it's apparent that this number is per pool instead of a total?

pivotal-jamie-klassen: In the foreseeable future we will might switch to  having many connection pools (one for scheduling, one for GC, one for API calls, etc) so documenting the specific number of pools will rot -- so maybe we should actually just provide a single knob for the max total number of connections that the web node can consume and dynamically allocate that number across multiple pools. I have heard @vito allude to something like this. The other option is that we should converge on all access to the database being brokered by the API, which basically means only one (API) pool per node, with the scheduler or garbage collector actually making API calls instead of hitting the DB directly. This is an approach I have heard @pivotal-jwinters favour explicitly.

PR #255, opened 2019-09-30
More cred manager docs

vito:


go/docs/plugin.go
```diff
@@ -78,6 +78,10 @@ func (p Plugin) QuickStart(content booklit.Content) booklit.Content {
 }
 
 func (p *Plugin) DownloadLinks() (booklit.Content, error) {
+	if os.Getenv("GITHUB_TOKEN") == "" {
+		return booklit.Empty, nil
```
pivotal-jamie-klassen: i'd rather have this break when misconfigured, say, on CI. But I like that it makes local builds faster/simpler/deterministic.

PR #261, opened 2019-10-14
Doc change for PR#4188.

evanchaoli:
This is the doc change for concurse/concourse#4188.

Signed-off-by: Chao Li <chaol@vmware.com>

### DIFF
--- a/lit/docs/jobs.lit
+++ b/lit/docs/jobs.lit
@@ -67,11 +67,24 @@ configured job consists of the following fields:
     \italic{Optional.} Keep logs for the last specified number of builds.
   }{job-build-log-retention-builds}
 
+  \define-attribute{minimum_succeeded_builds: number}{
+    \italic{Optional.} Keep a minimum number of successful build logs that
+    would normally be reaped. Requires \reference{job-build-log-retention-builds}
+    to be set to an integer higher than 0 in order to work. For example, if
+    \reference{job-build-log-retention-builds} is set to 5, and this attribute
+    to 1, say a job has the following build history:
+    7(f), 6(f), 5(f), 4(f), 3(f), 2(f), 1(s),
+    where f means failed and s means succeeded, then builds 2
+    and 3 will be reaped, because it retains 5 build logs, and at least 1
+    succeeded build log. Default is 0.
+  }{job-build-log-retention-minimum-succeeded-builds}
+
   Builds which are not retained by one of the above configurations will have
   their logs reaped.
 
   The following example will keep logs for any builds that have completed in
-  the last 2 days, while also keeping the last 1000 builds.
+  the last 2 days, while also keeping the last 1000 builds and at least 1
+  succeeded build.
 
   \codeblock{yaml}{{{
   jobs:
@@ -79,6 +92,7 @@ configured job consists of the following fields:
     build_log_retention:
       days: 2
       builds: 1000
+      minimum_succeeded_builds: 1
     plan:
     - get: 10m
     - task: smoke-tests


lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
...
 
+  \define-attribute{minimum_succeeded_builds: number}{
+    \italic{Optional.} Keep at lease specified succeeded builds. For example,
+    if specifies \reference{job-build-log-retention-builds} to 5, and this
+    attribute to 1, say a job has a build history: 7(f), 6(f), 5(f), 4(f),
+    3(f), 2(f), 1(s), where f means failed and s means succeeded, then builds 2
+    and 3 will be reaped, because it retains 5 build logs, and at lease 1
+    succeeded build log. Default is 0.
+  }{job-build-log-retention-minimum-succeeded-builds}
+
```
deniseyu: I wasn't involved at all with the original PR (which is probably a good thing now - fresh eyes and all  👀 ) so this paragraph reads a bit confusing to me! Going to suggest a multi-line rewrite, hopefully this works...

lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
     \italic{Optional.} Keep logs for the last specified number of builds.
   }{job-build-log-retention-builds}
 
+  \define-attribute{minimum_succeeded_builds: number}{
+    \italic{Optional.} Keep at lease specified succeeded builds. For example,
```
deniseyu:
```suggestion
    \italic{Optional.} Keep a minimum number of successful build logs that would normally be reaped. 
    Requires \reference{job-build-log-retention-builds} to be set to an integer higher than 0 in order to work.
```

lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
...
+    if specifies \reference{job-build-log-retention-builds} to 5, and this
```
deniseyu:
```suggestion
    For example, if \reference{job-build-log-retention-builds} is set to 5, and this
```

lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
...
+    attribute to 1, say a job has a build history: 7(f), 6(f), 5(f), 4(f),
```
deniseyu:
```suggestion
    attribute to 1, say a job has the following build history: 
    7(f), 6(f), 5(f), 4(f), 3(f), 2(f), 1(s),
```

lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
...
+    3(f), 2(f), 1(s), where f means failed and s means succeeded, then builds 2
```
deniseyu:
```suggestion
    where f means failed and s means succeeded, then builds 2
```

lit/docs/jobs.lit
deniseyu: (btw I know this line break won't render this way in the docs, but it makes it easier to reason about in Markdown)

lit/docs/jobs.lit
```diff
@@ -67,18 +67,29 @@ configured job consists of the following fields:
...
+    and 3 will be reaped, because it retains 5 build logs, and at lease 1
```
deniseyu:
```suggestion
    and 3 will be reaped, because it retains 5 build logs, and at least 1
```

evanchaoli:
@deniseyu Thanks for polishing the wording. Please help review again. This feature has been marked with release 5.7.

PR #264, opened 2019-10-16
Replaced the confusing request word with stdin payload for resource type implementation

gabipetrovay:
The term "request" is not mentioned anywhere else (that a "request" is made to the `in`, `out`, or `check` scripts). Hence calling the stdin payload a "request body" or just "request" is not uniform and confusing. On the other hand the documentation just mentions that this data is passed to the stdin of the corresponding scripts. Hence I called it "stdin payload".

### DIFF
--- a/go/docs/plugin.go
+++ b/go/docs/plugin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


vito:
Not sure what happened here but the commit range is empty now. 🤔 Sorry this didn't get any attention. Gonna close for now but feel free to re-submit.

PR #266, opened 2019-10-21
Change duplicated port number

rikkuness:
I am assuming this is the correct ports here? Docs talk about 7788 and 7788, where I assume should be 7777 and 7788?


pivotal-jamie-klassen:
@rikkuness good catch! You are definitely correct, but can you make sure your commit has a `Signed-off-by` line that matches the commit author? This is how you signify your agreement to our [DCO](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#signing-your-work).

rikkuness:
Yeah my bad, I edited it in web because laziness. I'll fix it 😆

Issue #268, opened 2019-10-25
Is 'on_error' supported as a Job attribute?


kdvolder:
What I mean is, there are step hooks like `on_failure`, `on_abort`, `on_success` and `on_error`.

Step hooks `on_failure`, `on_abort`, `on_success` are listed in the docs as being also acceptable as attributes on a Job directly (bottom of this page: https://concourse-ci.org/jobs.html).

The `on_error` step hook is not mentioned there. Is this an oversight in the docs?

Related issue: https://github.com/spring-projects/sts4/issues/383

pivotal-jamie-klassen:
job configs can definitely have an `on_error`: https://github.com/concourse/concourse/blob/master/atc/job_config.go#L18, so this is a gap in the docs! Would you consider making a PR to https://github.com/concourse/docs/blob/master/lit/docs/jobs.lit?

pivotal-jamie-klassen:
fixed in #269

PR #269, opened 2019-10-25
Add info about  'on_error' as a Job attribute

kdvolder:
See: https://github.com/concourse/docs/issues/268

### DIFF
--- a/lit/docs/jobs.lit
+++ b/lit/docs/jobs.lit
@@ -173,6 +173,11 @@ configured job consists of the following fields:
   \reference{on_failure} step attribute.
 }{job-on-failure}
 
+\define-attribute{on_error: step}{
+  \italic{Optional.} Step to execute when the job errors. Equivalent to the
+  \reference{on_error} step attribute.
+}{job-on-error}
+
 \define-attribute{on_abort: step}{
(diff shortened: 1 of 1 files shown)


pivotal-jamie-klassen:
@kdvolder thanks for this! do you mind signing off your commit? The check requires your commit to have a `Signed-off-by` line that matches the commit author, which signifies your agreement to our [DCO](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#signing-your-work)

kdvolder:
I tried, to fix the sign-off, looks  like it worked. 

For future reference, I feel like it is probably more efficient if the maintainers of the docs (i.e. you) make these kinds of small fixes themselves. 

Please consider that, for me to make this small change I have to first fork the repo, create a branch, do some edits, commit them, push them, go back to github and create a pull-request. Then on top of that I also had to figure out the sign-off warning at the end. It is a lot of red-tape for such a small edit.

Seems like if you just made the change yourself it would be more welcoming to people like me reporting little issues.

Anyhow, please don't get me wrong, I'm happy to help, and you are probably busy with other things than editing the concourse docs, but aren't we all?

Issue #267, opened 2019-10-25
Explain the difference between `on_failure`, `on_error`, `on_abort`


kdvolder:
The docs I'd like to see elaborated a bit are here: 

https://concourse-ci.org/on-failure-step-hook.html
https://concourse-ci.org/on-error-step-hook.html
https://concourse-ci.org/on-abort-step-hook.htm

It is somewhat unclear to me what the difference is between these 3. It would be nice to precisely explain under what circumstances a step is considered as 'aborted' versus 'failed' versus 'errored'. It seems to be that all 3 are somehow similar in that they all trigger when a step stopped in a 'abnormal' way. The difference between the 3 however is really just a bit of guessing. 

If I were to guess then I'd say that 'error' and 'failure' are synonyms. But 'abort' happens when the user stops the job/task from the ui.

Is this guess correct? I really have no idea, which is exactly why I think the docs could do with some further explanation.

pivotal-jamie-klassen:
I just looked at
https://github.com/concourse/concourse/blob/master/atc/config.go/#L320-L330 and I see slightly better prose descriptions in the comments.

marco-m:
hello @kdvolder :-)

the names have a 1-to-1 mapping with the Concourse UI that appears (and then disappears after a few seconds) if you move the mouse inside a pipeline view. See for example the public https://ci.concourse-ci.org/teams/examples/pipelines/job. I am reporting below a screenshot.

My way to remember the difference is by the color.

- failed -- on_failure -- red color: if the user script/task returned a non-zero exit code. Classic example failed compilation or failed test. Can also happen on a resource, for example if the resource returns a syntax error in how it has been configured or returns an external API error. This is definitive on a given build.
- errored -- on_error -- orange: if something went wrong at Concourse level. For example: if there are no more workers (temporary) to run a task or resource.
- aborted -- on_abort -- brown: if, as you mention, the user aborted the job.

<img width="182" alt="legend" src="https://user-images.githubusercontent.com/3075069/67597611-e52cf180-f76b-11e9-8ad1-5f9b4f5eca0a.png">

pivotal-jamie-klassen:
OK, so I tried to write up a comprehensive list of scenarios corroborating @marco-m's observations, but I realized it's a pretty big list. What I came up with so far:
[...]

PR #270, opened 2019-10-28
differentiate the 'bad' step hooks

pivotal-jamie-klassen:
Fixes concourse/docs#267.

I have tried to make these descriptions helpful and specific without being misleading. We could probably do a bit better giving users a sense of how build execution works overall, but I think this is a step in the right direction.


pivotal-jamie-klassen:
@kdvolder you are not a collaborator, but I am definitely interested in your feedback here.

Issue #254, opened 2019-09-17
Document resource pinning


pivotal-jamie-klassen:
As v5.6.0 is about to release, we have pretty well-defined fly and web workflows for resource pinning. I think the only real public-facing prose on the topic is [my blog post](https://medium.com/concourse-ci/resource-page-explained-eb99cf256fb5), so let's put some description of what it is and what it's for in the actual docs!

pivotal-jamie-klassen:
Done in #271

PR #271, opened 2019-10-29
document pinning versions

AnianZ:
As mentioned in #254 this should document how to pin a resource version using `fly` and the UI. I tried to follow similar pieces of documentation for that section but I have no problem structuring it differently or moving the page somewhere else if needed.

### DIFF
--- a/lit/docs/resources/managing.lit
+++ b/lit/docs/resources/managing.lit
@@ -28,3 +28,42 @@
   \italic{bosh-io-release} resource might use \code{version:11.2} in place of
   \code{ref:abcdef}.
 }
+
+\section{
+  \title{\code{fly pin-resource}}{fly-pin-resource}
+
+  To pin a resource to a specific version of that resource, run:
+
+  \codeblock{bash}{{{
+  $ fly -t example pin-resource --resource my-pipeline/my-resource \
+      --version ref:bceaf
+  }}}
+
+  Note that the version needs to be provided as key-value pairs. For the git
+  resource the \code{ref:} prefix is used. For example, the \italic{registry} 
+  resource might use \code{digest} as a prefix like \code{digest:sha256:94be7d7b}.
+
+  A comment can be provided using the \code{--comment} flag,
+  which is then also visible in the UI :
+
+  \codeblock{bash}{{{
+  $ fly -t example pin-resource --resource my-pipeline/my-resource \
+      --version ref:abcdef \
+      --comment "Some reason"
+  }}}
+
+  This can for example be used to pull in a fixed version of an external dependency
+  which might break your build in a new release. After the problem has been resolved
+  the pin can then be removed again.
+  Another example could be running a build with a set of older inputs when needed.
+
+  To remove the pin on a resource use:
+
+  \codeblock{bash}{{
+  $ fly -t example unpin-resource --resource my-pipeline/my-resource
+  }}
+
+  You can also pin a version via the UI by clicking on the pin icon next to the version on
+  the resouce page of the resource you want to change. A default comment is autmatically
+  generated containing your username and a timestamp which can be further edited.
+}
--- a/lit/docs/resources/managing.lit
+++ b/lit/docs/resources/managing.lit
@@ -52,9 +52,9 @@
       --comment "Some reason"
   }}}
 
-  This can for example be used to pull in a fixed version of an external dependency
-  which might break your build in a new release. After the problem has been resolved
-  the pin can then be removed again.
+  This can, for example, be used to pull in a fixed version of an external dependency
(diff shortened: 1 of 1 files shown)


lit/docs/resources/managing.lit
```diff
@@ -28,3 +28,42 @@
...
+  the pin can then be removed again.
+  Another example could be running a build with a set of older inputs when needed.
+
+  To remove the pin on a resource use:
+
+  \codeblock{bash}{{
+  $ fly -t example unpin-resource --resource my-pipeline/my-resource
+  }}
+
+  You can also pin a version via the UI by clicking on the pin icon next to the version on
```
pivotal-jamie-klassen: in the code, 'pin icon' means the icon in the top bar, but the we call the icon on each version a 'button'.
```suggestion
  You can also pin a resource via the UI by clicking on the pin button next to the desired version on
```

lit/docs/resources/managing.lit
```diff
@@ -28,3 +28,42 @@
...
+
+  \codeblock{bash}{{{
+  $ fly -t example pin-resource --resource my-pipeline/my-resource \
+      --version ref:abcdef \
+      --comment "Some reason"
+  }}}
+
+  This can for example be used to pull in a fixed version of an external dependency
+  which might break your build in a new release. After the problem has been resolved
+  the pin can then be removed again.
```
pivotal-jamie-klassen: i got confused because i thought this meant the pin had already been removed once
```suggestion
  the pin can be removed.
```

lit/docs/resources/managing.lit
```diff
@@ -28,3 +28,42 @@
...
+  which is then also visible in the UI :
+
+  \codeblock{bash}{{{
+  $ fly -t example pin-resource --resource my-pipeline/my-resource \
+      --version ref:abcdef \
+      --comment "Some reason"
+  }}}
+
+  This can for example be used to pull in a fixed version of an external dependency
+  which might break your build in a new release. After the problem has been resolved
```
pivotal-jamie-klassen: nit: commas
```suggestion
  which might break your build in a new release. After the problem has been resolved,
```

lit/docs/resources/managing.lit
```diff
@@ -28,3 +28,42 @@
...
+  the pin can then be removed again.
+  Another example could be running a build with a set of older inputs when needed.
```
pivotal-jamie-klassen: once [rerunning](https://github.com/concourse/concourse/projects/24) is delivered in v6.0, you won't need to use pinning this way!

lit/docs/resources/managing.lit
```diff
@@ -28,3 +28,42 @@
...
+  To pin a resource to a specific version of that resource, run:
+
+  \codeblock{bash}{{{
+  $ fly -t example pin-resource --resource my-pipeline/my-resource \
+      --version ref:bceaf
+  }}}
+
+  Note that the version needs to be provided as key-value pairs. For the git
+  resource the \code{ref:} prefix is used. For example, the \italic{registry} 
+  resource might use \code{digest} as a prefix like \code{digest:sha256:94be7d7b}.
```
pivotal-jamie-klassen: I think there might be a simpler way to say this (because it's a really good point), but I'm ok with it as is.

lit/docs/resources/managing.lit
AnianZ: Nice! I've been looking forward to that feature for a while. Should I remove the sentence or will v6.0 still take a while?

lit/docs/resources/managing.lit
AnianZ: I know :thinking: I tried to re-formulate it a little but didn't have a better idea either. I could copy the sentence from line 27 or leave it out as well, but for me personally the syntax of versions was not very clear from that. So I tried to explain it here a little differntly.

pivotal-jamie-klassen: don't sweat it, it's already an improvement 😄. I might take a crack at rephrasing in a follow-on PR.

lit/docs/resources/managing.lit
pivotal-jamie-klassen: we're playing around with releasing a beta - I think this is good to have, I'm almost just making a note to self to update the docs once 6.0 ships.

AnianZ: I'll make a note as well and let's see who remembers when 6.0 ships x)

AnianZ:
> This looks great, and ready to accept, but your commits need a sign-off! I'm guessing you applied those changes in the web UI, which often bites me.

Yes. Sorry for the holdup. I was not online at the weekend as well. But now I rebased on `master` and amended a sign-off to every commit.

PR #275, opened 2019-11-11
docs: add tracing section

cirocosta:
Hey,

This PR adds a minimal documentation for concourse/concourse#4607

Screen Shot 2019-11-12 at 8 37 47 AM

<details>
<summary>
see previous rendered version
</summary>
Screen Shot 2019-11-11 at 12 15 50 PM
</details>

wdyt?

thanks!


cirocosta:
>  so you actually need to specify the collector endpoint, including path components, like http://localhost:14268/api/traces - is that correct?

yeaah!

> Would you mind giving an example value for the CONCOURSE_TRACING_JAEGER_ENDPOINT variable?

suure, that's a great idea 👍

Issue #277, opened 2019-11-20
Document list of assignable roles and actions for RBAC


deniseyu:
We don't have user-facing documentation about the permissions that can be assigned to user types in Concourse. The roles can be `grep`'d in our code by looking at [this file](https://github.com/concourse/concourse/pull/4657/files#diff-806deacbe34acb68da0c0632773e6392) but that's not a very good user experience. We should clearly write out the strings used to indicate permissions on our [docs page about user roles & permissions](https://concourse-ci.org/user-roles.html), along with maybe some expansion of what each role can do if it's ambiguous.

Previously, this was less important to document because roles were hardcoded to certain permission sets internally, but after [this PR which made RBAC mappings configurable](https://github.com/concourse/concourse/pull/4657/files), it's more pertinent to tell our users how they can configure RBAC.

PR #282, opened 2019-12-12
document configurable RBAC

pivotal-jamie-klassen:
Fixes concourse/docs#277

### DIFF
--- a/lit/docs/auth/roles.lit
+++ b/lit/docs/auth/roles.lit
@@ -36,7 +36,8 @@ and viewers cannot.
   \reference{fly-set-team}, \reference{fly-destroy-team},
   \reference{fly-rename-team}, etc.
 
-  Admins also have permission to perform any of the actions below on any team.
+  Admins always have permission to perform any action on any team. You cannot
+  assign actions to the admin role using the \code{--config-rbac} flag.
 }
 
 \section{
--- a/lit/docs/auth/roles.lit
+++ b/lit/docs/auth/roles.lit
@@ -310,6 +310,8 @@ and viewers cannot.
   - OrderPipelines
   }}}
 
-  Please avoid specifying the same action under multiple roles in this file -
-  it can have unpredictable results.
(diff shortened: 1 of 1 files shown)


lit/docs/auth/roles.lit
```diff
@@ -43,77 +104,185 @@ Concourse comes with five roles: Concourse Admin, Team Owner, Team Member, Pipel
...
-    \table-row{                   }{Pause Job               }{✓    }{✓           }{✓            }{✓                       }{✘            }
-    \table-row{                   }{Trigger new Build       }{✓    }{✓           }{✓            }{✓                       }{✘            }
-    \table-row{                   }{Build History           }{✓    }{✓           }{✓            }{✓                       }{✓            }
-    \table-row{                   }{Paginate (<- ->)        }{✓    }{✓           }{✓            }{✓                       }{✓            }
-  }
+\section{
+  \title{Configuring RBAC}{configuring-rbac}
+  \warn{Configuring RBAC is \bold{experimental}, and this may change in the future.}
+
+  It is possible to modify the set of permitted actions for a given role, by
```
deniseyu: There's a quite subtle behavior that's hard to summarize with one word, but I think using "modify" here should come with a caveat. This sentence suggests to me that I can add or remove roles in a granular way, and someone coming from using GCP or AWS IAM might carry over assumptions that Concourse roles work the same way. With that mental reference point, I would be surprised to discover that in Concourse, it's not true (for example) that I can grant a role to a Team Viewer but not a Pipeline Operator.

I wonder if we can make it somehow clearer that Concourse roles work like a pyramid or a cascade, and _not_ like buckets of roles, like how other systems commonly work. Something like.. "Additional roles can be granted to a user type if they're not included by default; otherwise a role granted by default to a lower-permissioned role type can be "hoisted up" into the subset of roles belonging to a more highly-permissioned user type."

deniseyu: Also gonna tag in @matthewpereira for wordsmithing input!

PR #290, opened 2020-01-20
Update section on copying resource directories

RobertLK:
This advice was out of date and confusing, since Concourse now supports overlapping input and output https://github.com/concourse/concourse/issues/1799 . Update to recommend in-place update of files.


vito:
Some files move around so this became unmergeable. I've reflected the update on `master` myself. Thanks!

PR #292, opened 2020-01-23
Document Vault lookup paths

agurney:
See https://github.com/concourse/concourse/pull/5013.

This adds documentation for the new templating of Vault lookup paths, in standalone credential manager configuration and in `var_sources`. Additionally, this documents the Vault Enterprise "namespace" feature, which was not previously mentioned. It organizes the various ways of tweaking secret lookup into a consolidated account that hopefully makes sense for readers.


vito:
Thanks! This got a little stale with some docs shuffling, but I've updated it and merged manually. 👍

PR #293, opened 2020-01-23
Add docs and new site design for v6.0

clarafu:


less/base.less
```diff
@@ -122,7 +122,7 @@ h3 {
 }
 
 h2, h3 {
-  margin: 2rem 0 .8rem;
+  margin: 3rem 0 .8rem;
```
cirocosta: Should these be in a separate PR? 

Feels a bit odd to have styling coupled with all of the content 😁

lit/project/architecture.lit
```diff
@@ -1,78 +0,0 @@
-\title{Architecture}{architecture}
```
cirocosta: It seems like the removal of this file is preventing the rendering of `/`:

```
failed to load section: stat lit/project/architecture.lit: no such file or directory
```

I guess that's because it's referenced in few other places 🤔  

```
ag -Q "architecture"                                                                        [0/54]
lit/docs/internals.lit
13:\inline-header{Basic architecture}
20:\tutorial-image{images/concourse_architecture.png}
71:  registering \reference{architecture-worker}{workers} with the
92:  \title{Workers\aux{: container runtime & cache management}}{architecture-worker}

lit/project/contribute.lit
47:  getting started guide. You can also get an overview of the Concourse architecture
48:  by reviewing some of the documentation under \link{Concourse Architecture}{architecture.html}.

lit/project.lit
79:\include-section{./project/architecture.lit}
```

less/base.less
clarafu: Sure! I'll do that

clarafu: I added a PR for these changes #298

vito:
FYI this branch/PR will from here on contain all of the new v6.0 site design stuff in addition to the new documentation content. It's pretty hard to do one without the other since the content is dependent on plugins which reflect the website design etc.

That'll make this PR more difficult to review but I don't think we need to be so reliant on PR review for this stuff anyway; merging aggressively and pushing the new site often will probably be more effective.

PR #308, opened 2020-04-07
Add new auth params

pivotal-jwinters:
- CONCOURSE_CLIENT_SECRET
- CONCOURSE_TSA_CLIENT_SECRET


cskwrd:
Sorry to bring this PR from the dead. I can't find any documentation on how these values should be generated. How do I go about changing these from the example values?

taylorsilva:
Would have to dig up the PR these were added in. I don't _think_ you need to set them yourself though. I recall that Concourse will generate them for you. Looking in the main repo is your best bet to find an answer.

PR #310, opened 2020-04-22
document concurrent request limit metrics

jamieklassen:
concourse/concourse#5421

We don't have a natural section in the docs to talk about the API, so hopefully
it's enough to just mention the relevant metrics.

### DIFF
--- a/lit/docs/operation/metrics.lit
+++ b/lit/docs/operation/metrics.lit
@@ -65,6 +65,12 @@ metrics pipeline.
     # HELP concourse_builds_succeeded_total Total number of Concourse builds succeeded.
     # TYPE concourse_builds_succeeded_total counter
     concourse_builds_succeeded_total 0
+    # HELP concourse_concurrent_requests Number of concurrent requests being served by endpoints that have a specified limit of concurrent requests.
+    # TYPE concourse_concurrent_requests gauge
+    concourse_concurrent_requests{action="ListAllJobs"} 0
+    # HELP concourse_concurrent_requests_limit_hit_total Total number of requests that failed to acquire the lock to be served.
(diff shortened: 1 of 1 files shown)


lit/docs/operation/metrics.lit
```diff
@@ -65,6 +65,12 @@ metrics pipeline.
     # HELP concourse_builds_succeeded_total Total number of Concourse builds succeeded.
     # TYPE concourse_builds_succeeded_total counter
     concourse_builds_succeeded_total 0
+    # HELP concourse_concurrent_requests Number of concurrent requests being served by endpoints that have a specified limit of concurrent requests.
+    # TYPE concourse_concurrent_requests gauge
+    concourse_concurrent_requests{action="ListAllJobs"} 0
+    # HELP concourse_concurrent_requests_limit_hit_total Total number of requests that failed to acquire the lock to be served.
```
xtreme-sameer-vohra, line 71: Would it be clearer to say `requests were not served as they were above the concurrency limit` rather than mention lock which is more about the implementation ?

pivotal-jamie-klassen: this is a good idea - i just copy-pasted the documentation on that metric in the code. The source of truth is [here](https://github.com/concourse/concourse/pull/5429/commits/fe9ed50ddb25af4c4abdb7619b88405752978a2c#diff-1f503716ff020ce98016ddc7e74eedbcR161) in the connected PR, so I'll link this comment over there so we don't lose it.

pivotal-jamie-klassen: https://github.com/concourse/concourse/pull/5429/commits/7a91fac7d64a94cdf611678a484b4cc8bac5453a

PR #313, opened 2020-05-04
Add documentation for metric concourse_tasks_waiting

HannesHasselbring:
See: https://github.com/concourse/concourse/pull/5448

### DIFF
--- a/lit/docs/operation/metrics.lit
+++ b/lit/docs/operation/metrics.lit
@@ -112,6 +112,9 @@ metrics pipeline.
     # TYPE concourse_locks_held gauge
     concourse_locks_held{type="Batch"} 0
     concourse_locks_held{type="DatabaseMigration"} 0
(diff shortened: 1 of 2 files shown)


lit/docs/operation/metrics.lit
```diff
@@ -112,6 +112,9 @@ metrics pipeline.
     # TYPE concourse_locks_held gauge
     concourse_locks_held{type="Batch"} 0
     concourse_locks_held{type="DatabaseMigration"} 0
+    # HELP concourse_tasks_waiting Number of Concourse tasks currently waiting. Only emitted when using the limit-acitve-tasks placement strategy.
```
pivotal-jamie-klassen, line 115: There's a typo here:
```suggestion
    # HELP concourse_tasks_waiting Number of Concourse tasks currently waiting. Only emitted when using the limit-active-tasks placement strategy.
```
But it's also ok to just say
```suggestion
    # HELP concourse_tasks_waiting Number of Concourse tasks currently waiting.
```

PR #314, opened 2020-05-05
fix php pipeline in docs

bergpb:
fix composer in php pipeline and rename project to `laravel-websockets-git`, the same name in github repository


vito:
thanks!

Issue #334, opened 2020-05-27
Describe distributed tracing operations


jamieklassen:
It might be useful to break down each of the "operation" value that might get emitted -- stuff like `build` vs `scanner.check` vs `algorithm` stuff. Particularly with concourse/concourse#5659, there will be a bunch of operations and it's important for operators to be able to better understand them.

PR #335, opened 2020-05-28
describe distributed tracing operations

jamieklassen:
fixes concourse/docs#334

I didn't delve into a ton of detail on each of the private functions under a resolver -- I can imagine those things being subject to change and not very helpful to non-experts.

### DIFF
--- a/lit/docs/operation/tracing.lit
+++ b/lit/docs/operation/tracing.lit
@@ -70,23 +70,6 @@ although support for other systems is planned to expand as the underlying SDK
           \code{groupResolver.Resolve} -- This is the juicy part of the algorithm,
           which deals with
           \reference{schema.step.get-step.passed}{\code{passed}} constraints.
-          \list{
-            \code{groupResolver.tryResolve}
-            \list{
-              \code{groupResolver.trySatisfyPassedConstraintsForInput}
-              \list{
-                \code{groupResolver.tryJobBuilds}
-                \list{
-                  \code{PaginatedBuilds.migrateLimit}
-                }{
-                  \code{groupResolver.tryBuildOutputs}
-                  \list{
-                    \code{VersionsDB.migrateSingle}
-                  }
-                }
-              }
-            }
-          }
         }{
           \code{pinnedResolver.Resolve} -- This operation is used to determine
           inputs when \reference{version-pinning} is at play.
--- a/lit/docs/operation/tracing.lit
+++ b/lit/docs/operation/tracing.lit
@@ -40,7 +40,7 @@ although support for other systems is planned to expand as the underlying SDK
   \title{What's emitted?}
 
   Below is a summary of the various operations that Concourse currently traces.
-  They are arranged like a call tree, so that for operation described, its
+  They are arranged like a call tree, so that for each operation described, its
   sub-operations are described indented immediately below.
 
   \list{
--- a/lit/docs/operation/tracing.lit
+++ b/lit/docs/operation/tracing.lit
@@ -71,12 +71,16 @@ although support for other systems is planned to expand as the underlying SDK
           which deals with
           \reference{schema.step.get-step.passed}{\code{passed}} constraints.
           \list{
-            \code{groupResolver.tryResolve} \list{
-              \code{groupResolver.trySatisfyPassedConstraintsForInput} \list{
-                \code{groupResolver.tryJobBuilds} \list{
+            \code{groupResolver.tryResolve}
+            \list{
+              \code{groupResolver.trySatisfyPassedConstraintsForInput}
+              \list{
+                \code{groupResolver.tryJobBuilds}
+                \list{
                   \code{PaginatedBuilds.migrateLimit}
                 }{
-                  \code{groupResolver.tryBuildOutputs} \list{
+                  \code{groupResolver.tryBuildOutputs}
+                  \list{
                     \code{VersionsDB.migrateSingle}
                   }
                 }
--- a/lit/docs/operation/tracing.lit
+++ b/lit/docs/operation/tracing.lit
@@ -35,3 +35,77 @@ although support for other systems is planned to expand as the underlying SDK
   the default location that the \code{gcloud} CLI expects, or from GCP's
   metadata server (if Concourse is deployed on GCP).
 }
+
+\section{
+  \title{What's emitted?}
+
+  Below is a summary of the various operations that Concourse currently traces.
+  They are arranged like a call tree, so that for operation described, its
+  sub-operations are described indented immediately below.
+
+  \list{
+    \code{scanner.Run} -- One of the \reference{resource-checker-components}
+    responsible for determining which checks need to be run.
+    \list{
+      \code{scanner.check} -- This operation simply represents inserting the
+      check in the database.
+    }
+  }{
+    \code{checker.Run} -- follows from \code{scanner.check} above and will
+    appear under the same trace. This is where the \reference{checker} is
+    actually executing a \reference{resource-check} script.
+  }{
+    \code{scheduler.Run} -- This represents one tick of the \reference{scheduler}.
+    \list{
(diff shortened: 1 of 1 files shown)


lit/docs/operation/tracing.lit
```diff
@@ -8,21 +8,104 @@ Tracing in Concourse enables the delivery of traces related to the internal
...
+  \link{\code{GOOGLE_APPLICATION_CREDENTIALS} environment variable}{https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable},
+  the default location that the \code{gcloud} CLI expects, or from GCP's
+  metadata server (if Concourse is deployed on GCP).
+}
+
+\section{
+  \title{What's emitted?}
+
+  Below is a summary of the various operations that Concourse currently traces.
+  They are arranged like a call tree, so that for operation described, its
```
aoldershaw, line 43:
```suggestion
  They are arranged like a call tree, so that for each operation described, its
```

lit/docs/operation/tracing.lit
```diff
@@ -8,21 +8,104 @@ Tracing in Concourse enables the delivery of traces related to the internal
...
+  sub-operations are described indented immediately below.
+
+  \list{
+    \code{scanner.Run} -- One of the \reference{resource-checker-components}
+    responsible for determining which checks need to be run.
+    \list{
+      \code{scanner.check} -- This operation simply represents inserting the
+      check in the database.
```
aoldershaw, line 51: Not related to the docs, but out of curiosity - do we need to emit this operation if we emit `checker.Run` (i.e. could `checker.Run` follow from `scanner.Run`? what information would we lose if we didn't emit it?)

lit/docs/operation/tracing.lit
```diff
@@ -8,21 +8,104 @@ Tracing in Concourse enables the delivery of traces related to the internal
...
+    }
+  }{
+    \code{checker.Run} -- follows from \code{scanner.check} above and will
+    appear under the same trace. This is where the \reference{checker} is
+    actually executing a \reference{resource-check} script.
+  }{
+    \code{scheduler.Run} -- This represents one tick of the \reference{scheduler}.
```
aoldershaw, line 58: I'm not seeing any `scheduler.Run` spans being emitted in Concourse by looking through the code. I guess my initial assumption was that "operation" always corresponded with "span", but maybe that's not the case (or I'm just not finding it in the codebase)

lit/docs/operation/tracing.lit
```diff
@@ -8,21 +8,104 @@ Tracing in Concourse enables the delivery of traces related to the internal
...
+        \list{
+          \code{individualResolver.Resolve} -- This is used to determine versions
+          input to a \reference{get-step} without
+          \reference{schema.step.get-step.passed}{\code{passed}} constraints.
+        }{
+          \code{groupResolver.Resolve} -- This is the juicy part of the algorithm,
+          which deals with
+          \reference{schema.step.get-step.passed}{\code{passed}} constraints.
+          \list{
+            \code{groupResolver.tryResolve} \list{
```
aoldershaw, line 74: Not sure if there's anything you can do about this (it might just be a Booklit quirk) but the font-size on this nested list seems just slightly off (`1em` vs the usual `0.9em` that `code` in a `p` tag gets), making it look slightly wonky (but really not terrible):

Screen Shot 2020-06-01 at 10 36 12 AM

...whereas if we can somehow move this list into the `p` that holds the text above:

Screen Shot 2020-06-01 at 10 39 05 AM

...it looks a little better

lit/docs/operation/tracing.lit
jamieklassen: This is a great question. To me, this is an important distinction - so much so that maybe the operation names should be changed to make it clear.

The scanner and checker are two communicating sequential processes and the `checks` table is the message queue between them. `scanner.run` is the 'send' whereas `checker.Run` is the 'receive', and these events can be separated in time - the degree to which they are separated seems to represent a kind of 'latency' golden signal of SRE, but I haven't found a good way to formalize this.

What do you think of this? Is it worth renaming `scanner.check` to something more like `scanner.createCheck` or `scanner.insertCheck`?

lit/docs/operation/tracing.lit
jamieklassen: this operation is added in concourse/concourse#5659 -- https://github.com/concourse/concourse/pull/5659/files#diff-6dc7abff58d8493866b9ad8cd4121f01R52.

jamieklassen: "operation" is a property of a span though, you're totally right about that.

lit/docs/operation/tracing.lit
jamieklassen: ...this looks easier said than done. @vito pls? maybe it makes sense to drop the descriptions of all those private functions under groupResolver.

lit/docs/operation/tracing.lit
jamieklassen: nice snag

lit/docs/operation/tracing.lit
aoldershaw: I think a name like `scanner.createCheck` or `scanner.insertCheck` would be a bit more intuitive, but the docs do explain what it represents so it may not be necessary. I also like the CSP framing, I hadn't thought of Concourse like that but it makes sense.

lit/docs/operation/tracing.lit
vito: Hmm this brings up an interesting point around sustainability with documenting traces. It feels like if we go too far they're likely to become out of date. Should we have a convention where public functions are documented but private ones aren't? That way I can more freely refactor the internal bits, while knowing that any changes to "public" spans will require a documentation update.

re: the formatting - the syntax here feels kind of odd because it starts as a paragraph but then leads to a `\list`, which is a block element that can't exist in a paragraph (i.e. you can't have a `<ul>` in a `<p>`).

I think if you break the `\list` onto its own line it might do what you want it to - Booklit should parse the `\code` line as its own paragraph, which I think will make the font size consistent, since I think our CSS specifies `p code { font-size: 0.9em }` or something.

aoldershaw: Ah good point - I didn't piece together that the lowercase spans were intended to be "private". I assume there's no distinction between "Public" and "private" operations from Jaeger's perspective, right?

jamieklassen: I might go a step further and suggest that private functions should not be traced, i.e. "if it's worth tracing, it's worth making part of the public interface"

aoldershaw: Out of curiosity, is there any parallel notion to "log level" for traces? I guess something similar could be achieved with "tags"

jamieklassen: the concept I've heard of is "sampling priority" - I've heard it mentioned in [opentracing best practices](https://opentracing.io/docs/best-practices/#setting-sampling-priority-before-the-trace-starts) and seen in [client libraries](https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go#L56-L57), but it's not actually part of the opentelemetry spec, it's just a common convention. https://github.com/jaegertracing/jaeger/issues/365 has some related discussion.

vito: @jamieklassen Hmm yeah I think that's going too far. I think traces are really valuable for helping diagnose problems during development, and in this case the private traces were instrumental (heh) in finding the root cause of an issue in Hush House.

To me, tracing is something that should have a very low barrier to entry, even if means we under-document them. (I have differing thoughts on logging, though. IMO we should be more careful about what we log and try to make them more human-friendly, while preferring tracing/metrics for actual troubleshooting.)

jamieklassen: ok i'll drop the private stuff, with the exception of `scanner.check` which I believe is quite important

PR #341, opened 2020-06-17
Describe team attribute for set_pipeine step

tlwr:
Adds documentation for https://github.com/concourse/concourse/pull/5729

Aside, the new docs are awesome, as is booklit

### DIFF
--- a/lit/docs/jobs/steps.lit
+++ b/lit/docs/jobs/steps.lit
@@ -679,6 +679,23 @@ by configuring \reference{schema.step.on_failure} or
         in multiple files, the value from a file that is passed later in the
         list will override the values from files earlier in the list.
       }
+
+      \optional-attribute{team}{identifier}{
+        By default the \code{set_pipeline} step sets pipelines for the
+        pipeline's \reference{teams}{team}.
+
+        The \code{team} attribute can be used to specify a specific team.
+
+        Only the \reference{main-team} is allowed to set another team's
+        pipeline.  Any team other than the \reference{main-team} using the
+        \code{team} attribute will error, unless they reference their own team.
+
+        \warn{
+          The \code{team} attribute was introduced in Concourse v6.4.0. It is
+          considered an \bold{experimental} feature until its associated
+          \link{RFC}{https://github.com/concourse/rfcs/pull/31} is resolved.
+        }
+      }
     }
   }{
     \schema-group{\code{load_var} step}{load-var-step}{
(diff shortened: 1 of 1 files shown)


lit/docs/jobs/steps.lit
```diff
@@ -679,6 +679,23 @@ by configuring \reference{schema.step.on_failure} or
...
+
+      \optional-attribute{team}{identifier}{
+        By default the \code{set_pipeline} step sets pipelines for the
+        pipeline's \reference{teams}{team}.
+
+        The \code{team} attribute can be used to specify a specific team.
+
+        Only the \reference{main-team} is allowed to set another team's
+        pipeline.  Any team other than the \reference{main-team} using the
+        \code{team} attribute will error.
```
tlwr, line 691:
```suggestion
        Only the \reference{main-team} is allowed to set another team's
        pipeline.  Any team other than the \reference{main-team} using the
        \code{team} attribute will error, unless they reference their own team.
```

lit/docs/jobs/steps.lit
```diff
@@ -679,6 +679,23 @@ by configuring \reference{schema.step.on_failure} or
...
+        The \code{team} attribute can be used to specify a specific team.
+
+        Only the \reference{main-team} is allowed to set another team's
+        pipeline.  Any team other than the \reference{main-team} using the
+        \code{team} attribute will error, unless they reference their own team.
+
+        \warn{
+          The \code{team} attribute was introduced in Concourse v6.4.0. It is
+          considered an \bold{experimental} feature until its associated
+          \link{RFC}{https://github.com/concourse/rfcs/pull/31} is resolved.
```
vito, line 696: Can we add scarier wording and a link to the Discussion?

lit/docs/jobs/steps.lit
```diff
@@ -679,6 +679,23 @@ by configuring \reference{schema.step.on_failure} or
         in multiple files, the value from a file that is passed later in the
         list will override the values from files earlier in the list.
       }
+
+      \optional-attribute{team}{identifier}{
+        By default the \code{set_pipeline} step sets pipelines for the
+        pipeline's \reference{teams}{team}.
```
vito, line 685: This sounds a little ambiguous; how about something like this?:

```suggestion
        By default, the \code{set_pipeline} step sets the pipeline within the same \reference{teams}{team} that is running the build.
```

lit/docs/jobs/steps.lit
```diff
@@ -679,6 +679,23 @@ by configuring \reference{schema.step.on_failure} or
...
+
+        The \code{team} attribute can be used to specify a specific team.
```
vito, line 687: I wonder if we could frame this as "the default value for this field is the team that is running the build"?

tlwr:
I think I've incorporated your suggestions in the latest commit

I hope it reads clearer now
image

PR #342, opened 2020-06-18
update AWS secrets manager description

jamieklassen:
* clarify that a region is always required, but authentication can be done
  either via an access/secret key or a session token.
* The example IAM policy was incorrect -- per
  https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_identity-based-policies.html#permissions_grant-limited-resources:
  > If you don’t care about the region or account that owns a secret, you must
  > specify a wildcard character * , not an empty field, for the region and
  > account ID number fields of the ARN.
* give tips on writing secrets in the AWS console
* describe the health check and warn about rate-limiting.

### DIFF
--- a/lit/docs/operation/creds/aws-secretsmanager.lit
+++ b/lit/docs/operation/creds/aws-secretsmanager.lit
@@ -19,7 +19,7 @@
   attempt to use environment variables or the instance credentials assigned to
   the instance.
 
-  The web nodes's configuration specifies the following:
+  The web node's configuration specifies the following:
 
   \define-attribute{aws-secretsmanager-access-key: string}{
     A valid AWS access key.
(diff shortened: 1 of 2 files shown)


lit/docs/operation/creds/aws-secretsmanager.lit
```diff
@@ -6,12 +6,20 @@
...
+    an access key and secret key, or a session token
+  }{
+    the AWS region that your parameters are stored within.
+  }
+
+  If no access key, secret key, or session token is provided, Concourse will
+  attempt to use environment variables or the instance credentials assigned to
+  the instance.
+
+  The web nodes's configuration specifies the following:
```
pivotal-izabela-gomes, line 22: Just a typo, I _think_ the possessive form of the plural noun would be: web nodes' configuration.

PR #347, opened 2020-07-06
Add doc for renaming pipeline resources

mouellet:
PR: https://github.com/concourse/concourse/pull/5833

### DIFF
--- a/lit/docs/resources.lit
+++ b/lit/docs/resources.lit
@@ -42,9 +42,9 @@ with the following schema.
   }
 
   \optional-attribute{old_name}{identifier}{
-    The old name of the resource. If configured, the history of old resource will be
+    The old name of the resource. If configured, the history of the old resource will be
     inherited to the new one. Once the pipeline is set, this field can be
-    removed as the builds have been transfered.
+    removed as the history has been transferred.
 
     \example-toggle{Renaming a resource}{
       This can be used to rename a resource without losing its history, like so:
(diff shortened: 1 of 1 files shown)


lit/docs/resources.lit
```diff
@@ -41,6 +41,34 @@ with the following schema.
     organization}{https://github.com/concourse?query=-resource}.
   }
 
+  \optional-attribute{old_name}{identifier}{
+    The old name of the resource. If configured, the history of old resource will be
+    inherited to the new one. Once the pipeline is set, this field can be
+    removed as the builds have been transfered.
```
aoldershaw, line 47:
```suggestion
    The old name of the resource. If configured, the history of the old resource will be
    inherited to the new one. Once the pipeline is set, this field can be
    removed as the history has been transferred.
```

lit/docs/resources.lit
```diff
@@ -41,6 +41,34 @@ with the following schema.
...
+      \codeblock{yaml}{{
+      resources:
+      - name: new-name
+        old_name: current-name
+        type: git
+        source: {uri: "https://github.com/vito/booklit"}
+      }}
+
+      After the pipeline is set, because the builds have been inherited, the resource can
+      have the field removed:
```
aoldershaw, line 61:
```suggestion
      After the pipeline is set, the resource was successfully renamed, so the `old_name` field can
      be removed from the resource:
```

PR #351, opened 2020-07-15
update internal docs for TSA commands

vito:
per https://github.com/concourse/concourse/issues/2774#issuecomment-658931603


vito:
updated the diagram! merging now 👍

PR #353, opened 2020-07-22
Add doc for valid 'schema.identifier'

mouellet:
PR: concourse/concourse#5864

### DIFF
--- a/lit/docs/config/basic-schemas.lit
+++ b/lit/docs/config/basic-schemas.lit
@@ -61,22 +61,31 @@ just here for completeness!
 }
 
 \schema{identifier}{
-  An \italic{identifier} is string value. Currently, there are no validations
-  for what makes a string a valid identifier.
+  An \italic{identifier} is a string value. The following defines the allowed
+  character set for an \italic{identifier}:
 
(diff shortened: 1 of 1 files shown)


lit/docs/config/basic-schemas.lit
```diff
@@ -61,22 +61,30 @@ just here for completeness!
 }
 
 \schema{identifier}{
-  An \italic{identifier} is string value. Currently, there are no validations
-  for what makes a string a valid identifier.
+  An \italic{identifier} is string value. Currently, the validation will only
+  show as warnings. For the sake of future-proofing, you may want to conform to it.
```
aoldershaw, line 65: Could you move this comment about validation to the end (after you describe what the validation is)?

So, it'd be something like:

> An \italic{identifier} is a string value. The following defines the allowed character set for an \italic{identifier}:
> 
> ...
> 
> Currently, the validation will only show as warnings. For the sake of future-proofing, you may want to conform to it.

aoldershaw: Also, I know this was a problem before, but I just noticed the first sentence should be "An \italic{identifier} is **a** string value" (was missing the "a" before).

PR #354, opened 2020-07-22
Replace docker-image with registry-image

taylorsilva:
Since we plan on depracting docker-image and we'd prefer for people to
use and give feedback on registry-image resource instead.

I simply `ag`'d for `docker-image` and replaced all instances where it made sense to replace.

closes concourse/concourse#5678


vito:
Thanks!

PR #355, opened 2020-07-23
Updated the Nodejs example

kennedyalexander:
Using task inputs and outputs its possible to separate the npm install stage from the npm test stage.
This is useful as it allows the testing stage to run and hopefully only return error messages to do with testing and not something somewhat unrelated.
As this is an example it believe it should not resort to using exec commands as it skips over some of the more interesting features of Concourse.
This is my first PR I hope I got it right.

### DIFF
--- a/lit/examples/nodejs.lit
+++ b/lit/examples/nodejs.lit
@@ -5,6 +5,7 @@
 
 You can run the tests for a Nodejs application.
 
+This also demonstrates seperating the npm install stage from the test running.
 \frame{https://ci.concourse-ci.org/teams/examples/pipelines/nodejs}
 
 \section{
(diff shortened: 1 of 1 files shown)


lit/examples/nodejs.lit
```diff
@@ -5,43 +5,57 @@
...
-      - get: nodejs.org-git
-        trigger: true
-      - task: run-tests
+      - get: node-image
+      - get: repo
+      - task: install
+        image: node-image
         config:
+          inputs:
+          - name: repo
```
jamieklassen, line 38: Given how often you're using `source` elsewhere, this was probably what you were going for
```suggestion
          - name: repo
            path: source
```

lit/examples/nodejs.lit
```diff
@@ -5,43 +5,57 @@
...
+          run:
+            path: npm
+            args: ["install"]
+            dir: source
+      - task: test
+        image: node-image
+        config:
           inputs:
-            - name: nodejs.org-git
+          - name: repo
```
jamieklassen, line 51:
```suggestion
          - name: repo
            path: source
```

lit/examples/nodejs.lit
```diff
@@ -5,43 +5,57 @@
...
-            args:
-              - -c
-              - |
-                echo "Node Version: $(node --version)"
-                echo "NPM Version: $(npm --version)"
-                cd nodejs.org-git
-                npm install
-                npm test
+            path: npm
+            args: ["run", "test"]
```
jamieklassen, line 57: For something slightly simpler, you can do
```suggestion
            args: ["test"]
```
I'm not a nodejs expert -- is the `npm run test` invocation more idiomatic?

taylorsilva:
The content of everything under `lit/examples` is now pulled in from https://github.com/concourse/examples

I added and fixed your example with https://github.com/concourse/examples/commit/05181ee5756acdd2d1813f533d4b3269ed55b8fc

PR #357, opened 2020-08-05
document automatic archiving feature of set_pipeline step

### LINKED ISSUE #346: document set_pipeline step automatic archiving
https://github.com/concourse/concourse/issues/4873

chenbh:
Also remove the experimental warning for `set_pipeline` step

fixes #346

### DIFF
--- a/lit/docs/jobs/steps.lit
+++ b/lit/docs/jobs/steps.lit
@@ -566,12 +566,6 @@ by configuring \reference{schema.step.on_failure} or
   }{
     \schema-group{\code{set_pipeline} step}{set-pipeline-step}{
       \required-attribute{set_pipeline}{identifier}{
-        \warn{
-          The \code{set_pipeline} step was introduced in Concourse v5.8.0. It
-          is considered an \bold{experimental} feature until its associated
-          \link{RFC}{https://github.com/concourse/rfcs/pull/31} is resolved.
-        }
-
         Configures a \reference{pipelines}{pipeline}.
 
         The identifier specifies the name of the pipeline to configure. It will
--- a/lit/docs/jobs/steps.lit
+++ b/lit/docs/jobs/steps.lit
@@ -571,6 +571,22 @@ by configuring \reference{schema.step.on_failure} or
         The identifier specifies the name of the pipeline to configure. It will
         be configured within the current team and be created \italic{unpaused}.
 
+        Pipelines configured with the \code{set_pipeline} step are tied to the job that
+        configured them, and will be automatically archived in the following scenarios:
+        \list{
+          When the job runs a successful build which did not configure the pipeline (i.e.
+          the \code{set_pipeline} step was removed).
+        }{
+          When the job is removed from its pipeline configuration (see
+          \reference{schema.job.old_name} for renaming instead of removing).
+        }{
+          When the job's pipeline is archived or destroyed.
+        }
+        This means any job that uses \code{set_pipeline} should set all still-desired
+        pipelines in each build, rather than setting them one-by-one through many builds.
+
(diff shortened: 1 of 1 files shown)


lit/docs/jobs/steps.lit
```diff
@@ -566,17 +566,15 @@ by configuring \reference{schema.step.on_failure} or
...
-          \link{RFC}{https://github.com/concourse/rfcs/pull/31} is resolved.
-        }
-
         Configures a \reference{pipelines}{pipeline}.
 
         The identifier specifies the name of the pipeline to configure. It will
         be configured within the current team and be created \italic{unpaused}.
 
+        If a \code{set_pipeline} step is removed from a pipeline config (i.e. the step
+        or the job is deleted), any pipelines previously set by the step will
```
vito, line 575: Saying "pipelines" here seems slightly confusing, since the step should only correspond to a single pipeline. I guess that will change with the `across` step, since it'll technically be one step syntax, though many steps still run.

That being said, the semantics are slightly different for a *dynamic* `across` step - in this case, no step is removed, it just no longer runs with that particular set of instance vars. So even in the future, this will be kind of confusing. 😅 

Hmm... does this make sense?:

> Jobs which run the `set_pipeline` step become responsible for the lifecycle of the configured pipeline. When the job is removed, or when the job runs a build which succeeds without setting the pipeline, the pipeline will be automatically archived. This will happen if the step is removed from the build plan, or in the case of a dynamic `across` step which sets many pipelines instances, when the instance is no longer set.

No need to take this verbatim, I think it could use some tweaks, and that last part should probably be omitted since, y'know, the functionality doesn't exist yet. If you want, you could include it with Booklit comments (`{- these -}`).

Whaddaya think?

chenbh: I like the idea of explaining the internal logic while giving concrete examples.

How about:

> A job that successfully runs a `set_pipeline` step becomes the parent of the configured pipeline and is responsible for managing the child's lifecycle. The child pipeline will automatically be archived if the step is removed from the build plan. This can happen when the parent job runs a successful build that does not set the child pipeline, when the parent job is removed, or when the parent pipeline is deleted/archived.

vito: Getting there. I think the way that's worded is a little confusing, since it seems to imply that removing the parent job or parent pipeline removes it from the build plan.

Another subtle detail: it's not enough to remove it from the build plan; a build has to actually run. If we word it that way alone, someone might expect their pipelines to be archived at `fly set-pipeline` time when they remove it from the plan.

How about...:

> Pipelines configured with the `set_pipeline` step are tied to the job that configured them, and will be automatically archived in the following scenarios:
>
> * When the job runs a successful build which did not configure the pipeline (i.e. the `set_pipeline` step was removed).
> * When the job is removed from its pipeline configuration. (Use \reference{schema.job.old_name} for renaming jobs.)
> * When the job's pipeline is archived or destroyed.
>
> This means that jobs which use `set_pipeline` should set all still-desired pipelines in each build, rather than set them one-by-one through many builds.

lit/docs/jobs/steps.lit
```diff
@@ -566,17 +566,27 @@ by configuring \reference{schema.step.on_failure} or
...
+        \list{
+          When the job runs a successful build which did not configure the pipeline (i.e.
+          the \code{set_pipeline} step was removed).
+        }{
+          When the job is removed from its pipeline configuration (see
+          \reference{schema.job.old_name} for renaming instead of removing).
+        }{
+          When the job's pipeline is archived or destroyed.
+        }
+        This means any job that use \code{set_pipeline} should set all still-desired
```
vito, line 585:
```suggestion
        This means any job that uses \code{set_pipeline} should set all still-desired
```

PR #363, opened 2020-08-19
Update metrics for task waiting

taylorsilva:
* The tasks_waiting metric was updated from a Gauge to a Gauge Vector so
  labels could be included in the data
* A new metric was added, tasks_wait_duration, to show the elapsed time
  a task waited before execution

concourse/concourse#5981

### DIFF
--- a/lit/docs/operation/metrics.lit
+++ b/lit/docs/operation/metrics.lit
@@ -118,9 +118,24 @@ metrics pipeline.
     # TYPE concourse_locks_held gauge
     concourse_locks_held{type="Batch"} 0
     concourse_locks_held{type="DatabaseMigration"} 0
+    # HELP concourse_tasks_wait_duration Elapsed time waiting for execution
+    # TYPE concourse_tasks_wait_duration histogram
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="1"} 0
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="15"} 0
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="30"} 0
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="60"} 0
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="120"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="180"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="240"} 1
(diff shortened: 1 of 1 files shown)


lit/docs/operation/metrics.lit
```diff
@@ -118,9 +118,24 @@ metrics pipeline.
...
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="60"} 0
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="120"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="180"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="240"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="300"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="600"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="1200"} 1
+    concourse_tasks_wait_duration_bucket{platform="linux",teamId="1",workerTags="",le="+Inf"} 1
+    concourse_tasks_wait_duration_sum{platform="linux",teamId="1",workerTags=""} 60.0001929
+    concourse_tasks_wait_duration_count{platform="linux",teamId="1",workerTags=""} 1
```
aoldershaw, line 135: I'm confused - where does this information appear in the docs? All I see is:

Screen Shot 2020-08-19 at 10 05 56 AM

aoldershaw: To be fair, it seems consistent with how other metrics are defined, I'm just not sure how it works tbh

taylorsilva: +1 that. I didn't dig into it but I'm guessing someone (@vito ?) wrote a lit plugin to parse the output of the metrics endpoint. If you bring up concourse with the `./hack/overrides/prometheus.yml` and visit http://localhost:9100/ you'll see the output in this doc. That's where I got this "data" 😁

vito: Yep, all the metrics docs are generated via the `PrometheusDocs` plugin function, so all you need to do is copy paste the response. The response already includes descriptions and types so I figured this would be an easier flow than hand typing the same info in two places.

taylorsilva: One small issue with this that I haven't seen written anywhere (maybe it should be a comment in this file?) is that not all metrics are listed in the response. There are a few metrics (e.g. the two this PR adds/updates) that don't appear in the output until some data is generated for them. So it's not always as simple as "copy and paste the entire response" because the response you're copying probably doesn't contain all metrics 😬

vito: Ah dang that's true.

PR #366, opened 2020-09-02
fly: Document client certificate options

avanier:
This PR documents the client certificate options added by concourse/concourse#6017.


aoldershaw (approved):
Thanks!

PR #371, opened 2020-09-23
Clarify how the out script receives resources

compnski:
Use a similar wording to the in script to indicate it receives this as command line argument $1.


vito (approved):
Thanks for doing this! This does read more clearly.

PR #375, opened 2020-10-05
Update documentation regarding the support for Vault KV V2.

daviddob:
https://github.com/concourse/concourse/pull/6115


vito (approved):
thanks!

PR #379, opened 2020-10-26
add examples for nested fields in static vars

chenbh:
Realized that nested fields weren't fully documented


aoldershaw:
Nice, there's some documentation around the syntax of nested fields here: https://concourse-ci.org/vars.html#var-syntax. The examples are nice, though

chenbh:
Hmm, thanks for pointing that out, we should probably reference that as well

aoldershaw (approved):

PR #378, opened 2020-10-14
At least one job is required for a pipeline to be valid

taylorsilva:
Related change concourse/concourse#6159

Preview of this change:

image

PR #380, opened 2020-10-28
Temporarily revert OTLP docs

chenbh:
Since we're release 6.7.0 with only the resource type defaults, let's take out OTLP stuff for now.

There's also https://github.com/concourse/docs/pull/378, but while the feature in fly cli that enforces that won't make it into this release, it's enough of a good practice that we feel it'll be fine to leave it in


jamieklassen (approved):
lgtm 😢

PR #390, opened 2020-12-10
generate example pipelines from github

aoldershaw:
when we change the example pipelines, it's easy to forget (or just not
be bothered) to update the example in the docs.

this commit adds a new `remote-codeblock` command that generates a
codeblock by making an HTTP request to the provided endpoint. this lets
us point at the examples repo directly so the pipeline configs stay in
sync

Signed-off-by: Aidan Oldershaw <aoldershaw@pivotal.io>


muntac (changes requested):
Nifty! Looks like the Java example was never included in the examples [list](https://github.com/concourse/docs/blob/9cd62b3d93ab0a6f034a91cab84f2cd7e2dad3c9/lit/examples.lit#L27) even though there's an example file. It would be good to add it.

aoldershaw:
@muntac nice catch! fixed

muntac (approved):
LGTM!

PR #397, opened 2021-01-23
Fix incorrect information about team owner permissions

Krenair:
Team owners can rename and destroy their team.

Fixes https://github.com/concourse/concourse/issues/6422


clarafu (approved):
Thanks for the fix!

PR #402, opened 2021-01-27
Fix out of date info on resource checking

aoldershaw:
checks-as-builds made some things out of date

Signed-off-by: Aidan Oldershaw <aoldershaw@pivotal.io>


vito (approved):
looks good! nice touch integrating `check_every: never`.

PR #407, opened 2021-02-24
Document build log retention at the Concourse level

aoldershaw:
...and mention that build logs aren't reaped when pipeline/job is paused

### DIFF
--- a/lit/docs/install/web.lit
+++ b/lit/docs/install/web.lit
@@ -333,6 +333,47 @@ as performing all pipeline scheduling. It's basically the brain of Concourse.
   }
 
   \section{
+    \title{Build log retention}{build-log-retention}
+
+    Build logs are stored in the DB - if they are not cleanup up every once in
+    a while, the storage usage for build logs will continue to grow as more builds
+    run. While this is usually fine for small Concourse instances, as you scale up,
+    you may run into storage concerns.
+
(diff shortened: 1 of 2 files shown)


lit/docs/install/web.lit
```diff
@@ -332,6 +332,47 @@ as performing all pipeline scheduling. It's basically the brain of Concourse.
...
 
+  \section{
+    \title{Build log retention}{build-log-retention}
+
+    Build logs are stored in the DB - if they are not cleanup up every once in
+    a while, the storage usage for build logs will continue to grow as more builds
+    run. While this is usually fine for small Concourse instances, as you scale up,
+    you may run into storage concerns.
+
+    To clean up old build logs, Concourse periodically scans for builds whose
```
clarafu, line 343: nit: Wondering if it's worth making it clear that you have to configure Concourse to reap builds logs rather than it being automatically on?

```suggestion
    To clean up old build logs, you can configure Concourse to periodically scan for builds whose
```

clarafu (approved):
Very useful section to add to the docs! :D

lit/docs/install/web.lit
aoldershaw: good point, thanks! fixed

clarafu (approved):

PR #408, opened 2021-02-25
Add Clarification for Vault Paths with Many Levels

rcsalome:
This adds a tiny example of how to use the `/` separator to traverse a
Vault path. The intention is to make it clear that the `.` is the field
separator and should not be used as `((foo.bar.baz.quz))`.


clarafu (approved):
Thanks for the clarification! 😄

PR #410, opened 2021-03-18
add RFCs table to front page

vito:
related to (& dependent on, for some of the links) https://github.com/concourse/rfcs/pull/97

* adds a table of the top 10 most popular RFCs (by github reactions)
* call-to-action for feedback and deep-links to open questions within RFCs
* replaces the manual RFC round-up blog posts process; hopefully directs a lot more traffic to the RFCs

i'd include a screenshot but it's a little janky due to the fixed background image position - try it out with `./scripts/build -s 3000`!


go/docs/github.go
```diff
@@ -0,0 +1,218 @@
...
+			"FlyLinuxURL":     booklit.String(flyLinuxURL),
+			"FlyDarwinURL":    booklit.String(flyDarwinURL),
+			"FlyWindowsURL":   booklit.String(flyWindowsURL),
+		},
+	}, nil
+}
+
+func (p *Plugin) githubClient(ctx context.Context) *githubv4.Client {
+	var tc *http.Client
+	var githubToken = os.Getenv("GITHUB_TOKEN")
```
aoldershaw, line 207: Not much you can really do about this without complicating the code a fair bit, but kind of a bummer that this means you need to set `GITHUB_TOKEN` whenever you want to run the docs locally

vito: ah yeah true, let me change this to fall back on filler data when the token isn't set

vito: done!

vito:
rebased to split the community banner into https://github.com/concourse/docs/pull/411 (though these changes are still on top of it)

go/docs/github.go
```diff
@@ -0,0 +1,237 @@
...
+		if release.TagName == nil {
+			return nil, fmt.Errorf("no tag name for release %v", release)
+		}
+		version, err := semver.ParseTolerant(*release.TagName)
+		if err != nil {
+			return nil, err
+		}
+
+		if latestRelease == nil || version.GT(latestVersion) {
+			latestRelease = &release
```
taylorsilva, line 152: I kept getting 6.5.1 as my latest version. You're returning a pointer to `release` which gets updated after every loop.
image

go/docs/rfcs.go
```diff
@@ -0,0 +1,449 @@
...
+func (pr PullRequest) ByTotalReactions() int {
+	var s int
+	for _, reaction := range pr.Reactions {
+		s += reaction.Count
+	}
+
+	return s
+}
+
+func (pr PullRequest) ByOpenQuestions() int {
```
taylorsilva, line 54: This and other `By...()` appear to not be used anywhere. I'm fine keeping them, just want to make sure I'm reading it right that they're currently not being used.

taylorsilva (changes requested):

go/docs/rfcs.go
```diff
@@ -0,0 +1,449 @@
...
+			"https://github.com/%s/%s/blob/%s/%s",
+			repo.Owner.Login,
+			repo.Name,
+			node.HeadRefOid,
+			proposalPath,
+		)
+
+		questionsURL, totalQuestions, err := countQuestions(proposalURL)
+		if err != nil {
+			return nil, fmt.Errorf("count open questions: %w", err)
```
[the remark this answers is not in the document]
vito: Hahah, my own PR (https://github.com/concourse/rfcs/pull/97) nuked my other PR. Good catch!

go/docs/rfcs.go
vito: Yeah, these are here because I experimented with a few other renditions of the table and I wanted to be able to quickly switch between them from the Booklit document:

```tex
\rfcs-table{10}{ByTotalReactions}
```

Happy to remove them if you prefer.

go/docs/github.go
vito: fixed!

taylorsilva (commented):
Looks good!
image

Only thing that makes me go 🤨 is the way the number of open question are displayed. On first viewing I thought it was a font error, like it was trying to display an unknown character. I actually only understood what it meant because I have context on the format of RFC's. I don't think this will be obvious to people seeing it for the first time and not being familiar with our RFC format. Maybe @matthewpereira has a better idea of how to represent "number of open questions" in a succinct way? Maybe it should be it's own column called "Open Questions"?

Shouldn't block this pr, but I wanted to discuss it before approving.

taylorsilva:
Excuse the hacky formatting, playing around in the browser, adding open q's as a column
image

vito:
@taylorsilva What happened with the background image at the top there? 😆

And yeah I struggled with the open questions representation for a bit. It had its own column at one point, but I opted to bring it closer to the proposal since it's a sub-heading of it. Here's a cheap attempt that swaps the "?" for an acronym (yay), "OQ":

image

(Also that's what the header's supposed to look like - does it work for you or was that just a weird screenshot?)

taylorsilva:
lol yeah firefox messed up the screenshot! it looks fine for me, same as what you're seeing.

Gonna approve now, maybe someone will think of something better later.

taylorsilva (approved):

lit/index.lit
```diff
@@ -17,16 +17,16 @@
...
 
   \rfcs-table{10}{ByTotalReactions}
 
-  Help shape Concourse into a tool that meets your needs by \link{submitting
-  feedback}{https://github.com/concourse/rfcs/#providing-feedback-on-an-rfc}!
-  The above RFCs are ranked by their total GitHub reactions, with the aim being
-  to focus feedback on popular RFCs so that they reach resolution, making room
-  for more on the table.
+  Help shape Concourse into a tool that fits your needs by \link{submitting
+  feedback}{https://github.com/concourse/rfcs/#reviewing-rfcs} on the RFCs
```
taylorsilva, line 28: Initially I was like "what, this is wrong?". Then I remembered your open PR 😅

taylorsilva (approved):
LGTM

vito:
Thanks for the review! Just merged the RFC PR, so merging this no.

PR #412, opened 2021-03-23
Remove cc.xml example

icelynjennings:
Explicit URL `http://ci.concourse-ci.org/api/v1/teams/main/cc.xml` appears inaccessible with message `not authorized` - presumably to all but `ci.concourse-ci.org` staff. 

Happy to edit with a different example if suggested.


taylorsilva (approved):
This whole page needs a bit of love! Thanks for the PR :)

PR #418, opened 2021-04-09
Concourse tutorial

taylorsilva:
Looking for feedback on the structure, pacing, explanations, everything!

TODO:
- [ ] ~Passed constraints~ maybe put that in a guide


taylorsilva:
@muntac can't assign you so tagging so you get a github notification

taylorsilva:
Assigning myself to clear this out from the project board

taylorsilva:
This is ready for review. Nit-pick it away!

EstebanFS:
@aoldershaw

lit/docs/getting-started/resources.lit
```diff
@@ -0,0 +1,542 @@
...
+  That's a high-level overview of the resource interface. Next, we will learn
+  how to add resources to a pipeline.
+}
+
+\section{
+  \title{Get Steps}
+
+  Let's add some automation to our \code{hello-world.yml} pipeline and have
+  it \bold{trigger every one minute}. We will use the \link{time
+  resource}{https://github.com/concourse/time-resource/} to accomplish this.
```
aoldershaw, line 113: IMO the most easiest resource type to explain resources is the `git` resource, since most developers are familiar with git and the commit model to some extent. The `time` resource, on the other hand, feels a little less natural - sure, time *is* external state, but:
1. I suspect most people don't think of it in the same way as e.g. a `git` repo, and
2. We're currently working to get rid of it as a resource type (in favour of a var source), so I feel like introducing it as a resource only to change the mental model shortly thereafter is a bit confusing

So, maybe we could point to a sample git repository instead? I think that'd help people picture real-world use cases of Concourse automation. For explaining the `put` step, we could suggest they create a fork of the repo and push to that?

lit/docs/getting-started/resources.lit
```diff
@@ -0,0 +1,542 @@
...
+  Adding a \reference{resource-types}{resource type} to your pipeline looks
+  very similar to adding a resource. You can even override the bundled resource
+  types by re-declaring them in your pipeline.
+
+  Remember, a resource is a container image. So to pull in a new
+  resource type you need to tell Concourse \italic{where} to pull the image
+  from. This is done by using the built-in \link{registry-image
+  resource}{https://github.com/concourse/registry-image-resource/}.
+
+  Here is how we would re-declare the \code{time} resource in our
```
aoldershaw, line 479: This maybe isn't the most compelling example of why you'd want to use a custom resource type. Maybe we could use one from the community? There's a good example of using the RSS resource in https://concourse-ci.org/resource-types.html, which I think is a pretty intuitive example.

lit/docs/getting-started/resources.lit
```diff
@@ -0,0 +1,542 @@
...
+  \list{
+    \reference{how-to-guides} - Contains tips for writing pipelines and examples of common pipeline workflows
+  }{
+    Check out all the reference documentation:
+    \list{\reference{jobs}}{\reference{tasks}}{\reference{resources}}{\reference{resource-types}}
+  }{
+    \reference{implementing-resource-types}{Implement your own resource type}
+  }{
+    Find other resources at
+    \link{resource-types.concourse-ci.org}{https://resource-types.concourse-ci.org/}
```
aoldershaw, line 536: I think maybe we can call this out more explicitly in the previous section. Glancing at this section makes it seem like it's a summary, so readers may be inclined to skip over this section if they feel they understood the rest (since they might think there's no new information).

aoldershaw (commented):
A few small suggestions but looks really great! Nice work!

lit/docs/getting-started/resources.lit
```diff
@@ -0,0 +1,542 @@
...
+  things that resources can do:
+
+  \list{
+      You want something to run every five minutes? \link{Time resource}{https://github.com/concourse/time-resource/}.
+  }{
+      You want to run tests on every new commit to the main branch? \link{Git resource}{https://github.com/concourse/git-resource}.
+  }{
+      Run unit tests on new PR's? \link{Github-PR resource}{https://github.com/telia-oss/github-pr-resource}.
+  }{
+      Build the latest image of your app? \link{Registry-image resource}{https://github.com/concourse/registry-image-resource/}
```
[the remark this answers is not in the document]
taylorsilva: lol I was literally thinking "don't get this backwards" when writing it and I totally got it backwards 😆

lit/docs/getting-started/resources.lit
```diff
@@ -0,0 +1,542 @@
...
+
+  \codeblock{yaml}{{
+  resources:
+  - name: every-1min
+    type: time
+    source:
+      interval: 1m
+  # add a new resource
+  - name: timestamps
+    type: mock  # from https://github.com/concourse/mock-resource
```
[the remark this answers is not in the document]
taylorsilva: Yup, totally agree with you. The resource section here was the hardest to write and this section even more so. I might re-write this entire section to use this https://github.com/marco-m/concourse-in-a-box which simulates external stuff like s3

lit/docs/getting-started/resources.lit
taylorsilva: git would be my goto explanation as well. I just dislike the overhead of having to tell someone to setup a git repo 😓 
Anyway, I rewrote the entire section using git instead cause it is better 😁

taylorsilva:
@aoldershaw thanks for the review! Didn't surprise me that most comments were on the resource page. It was the hardest to write and it only went through one revision while all the other pages went through many more revisions 😅 

Changed the resource section to use the git resource, it does just work better for explaining resources to devs.

EstebanFS (approved):

PR #427, opened 2021-05-20
add boclips as user, after chat with Andrew Bruce

DanielJonesEB:
Had a chat with Andrew Bruce at Boclips via LinkedIn messages, who suggested it was cool to add Boclips to the list, and provided a URL to their Concourse instance: https://concourse.devboclips.net/


muntac:
CC: @taylorsilva

PR #435, opened 2021-06-03
update docker-compose.yml and document dns-proxy-enable and allow-host-access

chenbh:
We removed the env var from the docker image in https://github.com/concourse/concourse-docker/commit/a9a4a0c4cbc2dfee14525ab3caffaf258892da12


taylorsilva (approved):
LGTM - glad we went with that change to the docker image. Was weird having that always on

PR #437, opened 2021-06-15
Add CSP header to allow for iframing in tutorials

taylorsilva:


vito (approved):
Seems fine, but out of curiosity do you know what puts this in an iframe? Is this for local docs dev?

taylorsilva:
I added it as an iframe to the quickstart page and throughout the getting started tutorial.

PR #440, opened 2021-06-26
Quote the URL

bngsudheer:
Otherwise shell will interpret special characters such as & differently.


taylorsilva (approved):
Thanks for the PR!

PR #442, opened 2021-07-05
Add guide for templating with ytt

taylorsilva:
Rendered
image


taylorsilva:
@aoldershaw

taylorsilva:
Had the idea of spinning this into something more general but I don't have the motivation for doing this right now so I'm closing this. Thanks for the feedback though Aidan!

PR #462, opened 2021-09-24
Document the injection of the `TRACEPARENT` environment variable in `run` commands for context propagation

cyrille-leclerc:
Document the injection of the `TRACEPARENT` environment variable in `run` commands for context propagation


cyrille-leclerc:
Commit amended to add "Signed-of-by"

PR #466, opened 2021-10-07
Explain how dynamic vars work with the across step

taylorsilva:
and explain limits of the across step with get and put steps

image

and

image


keevan:
> I added a bunch of examples to the across step as well to hopefully better guide usage of it.

This is great, thanks a bunch :)

PR #484, opened 2022-02-03
Restructure docs to act as a basis

tomschr:
@taylorsilva 
This PR contains a lot of changes. Remember, in this PR I just focused on the structure. We should probably rework the content in a different PR.

Basically this PR is just about moving content from A to B. The content itself isn't modified apart from indenting or creating new files. Think about more of creating the basis for upcoming changes.

## Some notes about my changes

* Cleaned up the header on the first page so it contains only "beginners", "users", and "references".
* Prepared an admin topic. However, I haven't used it and it's currently disabled. We can decide later, if we really need it.
* For some parts I wasn't sure about if it's a reference or a concept. I moved it mostly to the reference part, but that would probably something that needs to be adapted.
* The fly command contains now a collection of subcommands, mostly moved from other parts.
* I've tried to make the changes step by step: move content, test and reload the page in the browser, commit, and repeat. I hope the commits make some sense and are easy to follow.
* If you prefer, we can squash the commits in the end.

## Benefits

* Makes the basis for a slimmer, user-focused, and well-organized documentation.
* The fly subcommands are sorted and are in separate pages. This makes it easier for referencing, searching, and reading.
* The header contains only 3 main topics (apart from "blog" and "discuss"). This (hopefully) drives our readers to the right slot.

## TODOs

As I said, it's only the beginning. :wink: If you think this is useful and you agree to proceed further, I'd recommend the following ideas:

* Beginners
  * Add some more content?
   The overall topic is a bit meager as the meat is in other parts. Maybe we could add something else that a beginner needs to know? If not, we can leave it as is.
  * Write the "What's next?" topic
    Currently this file is empty and is a placeholder. Here we could give our readers some hints what it's useful to read next.

* Users
  * Adapt content to make it more task-oriented
   The current content is a mixture of concepts and tasks. We should separate both. To have a more task-oriented look, we should add more procedures and move conceptual parts into another file.
  * Adapt the titles accordingly.
  * Correct the order, if needed.

* References:
  * Use the template file `lit/references/fly/_template.lit` for all fly subcommands.
    The idea is that it acts as, well, template for each and every subcommand of `fly`. It has a well-known manpage structure.
  * Add missing/change existing fly subcommand pages
    I count 75 subcommand of `fly`, but I see only 27 descriptions inside `lit/references/fly`. :wink:  The existing content should to be rewritten to follow the structure of the template. Furthermore, missing description of subcommands needs to be a added.
  * Write missing API references
  * Adapt order, rearrange topics
   The table of content (TOC) on the left sidebar is a bit long for my taste. I would like rearrange the existing topics into some parent-child relationships. For example, the TOC should show these main elements: "The fly CLI", "Workflow format", "API", "Internals" and maybe some else which doesn't belong.  The "Workflow format" would contain all the sub-topics like jobs, vars etc.

----

What do you think? Any wishes, any objections? Something you would like to change?


taylorsilva:
WOW! Thanks for the fast turnaround!

Before:
image

After:
image

I'm liking the structure this is taking. I'm not liking the words in the header though, specifically "beginners" and "users". The "users" header feels vague and I feel it's unclear what I would find by clicking on that item.

---

### Beginners
> Write the "What's next?" topic

This guide: https://concourse-ci.org/exploring-task-input-and-output-scenarios.html
and then this: http://localhost:3000/common-pipeline-practices.html
Would be my recommended next stops for a new user.

Maybe a future guide we could write would be about how to add/write a concourse pipeline for an existing git repo.
### Users

> Adapt content to make it more task-oriented

Sounds good to me 👍 

### References

> Use the template file lit/references/fly/_template.lit for all fly subcommands.

I like this. I know in the past a lot of things got shut down because "they're too hard to keep updated". From what I've seen though these commands don't really change besides flags being added (e.g. `--team` to various commands, and still ongoing).

> Write missing API references

We've avoided doing documenting the API because we don't want to support it, so this has been our way of discouraging usage of the API. We also want to discourage people from automating Concourse itself, but maybe that's a silly reason? Though this is how we've gotten feedback and created stuff like the `set_pipeline` step and eventually adding the `teams` field to that step as well.
wdyt?

> Adapt order, rearrange topics

Instead of "Workflow Format" I think "Pipeline Format" would be better. You don't see the word "workflow" pop-up anywhere in Concourse.

taylorsilva:
@clarafu @xtremerui Big docs change happening here, please drop any thoughts/feedback y'all have.

tomschr:
> WOW! Thanks for the fast turnaround!

Yes, I was also surprised. Sorry for the bigger change, but I wanted to show something substantial.

> I'm liking the structure this is taking.

Thanks. The idea was to clean up the header so it contains only the necessary bits.

> I'm not liking the words in the header though, specifically "beginners" and "users". The "users" header feels vague and I feel it's unclear what I would find by clicking on that item. Can't think of a better word atm though

We can change that, of course. They are just working titles. What about renaming "beginners" to "start"? For "users" I don't have something better ATM either.

taylorsilva:
> Another alternative would be to rename "users" to "using" and just collect only task-oriented topics. We could introduce another item between this and references and name it "concept". So the overall header would be like:
> ```
> Start   Using   Concepts   References
> ```

Would "Guides" work better? Was looking at the [k8s docs](https://kubernetes.io/docs/home/) and pulled "Guides" from there. "Using" feels like I'm saying half a sentence ("using" what?). Still feels task-oriented to me. 
```
Start   Guides   Concepts   References
```

Otherwise,

> ... just collect only task-oriented topics ... That would involve moving some topics, but I think, that should be ok. What do you think?

I'm good with this change, I like it 👍 Could have user-focused sub-sections within that section then, probably "admins" and "advanced users".

> Good point! Actually I was searching for a good term. As GitHub, GitLab, and others have similar terms, I would suggest to have some definitions, glossary, or terminology about all these terms. It's even more important if the scope of these terms are different.

A bunch of this stuff is loosely defined at the top of this page actually: https://concourse-ci.org/docs.html
which I admit is quite easy to miss all together! A page dedicated to Concourse-specific terms sounds like a good idea, maybe under References? Or Concepts?

> Regarding the "too hard to keep updated". Depends on how you develop Concourse. In an agile way, you need to add the task "documenting new commands in our doc" into the definition of done. Or you extract the documentation from some comments etc. inside the source code. Would be a bit cumbersome, but that way developers have code and documentation in one place. I have no idea if that's possible with this project, but I did some API documentation this way.

All flags for the fly and concourse cli are stored in struct tags, which means they should be somewhat easy to access and output to other formats. We can worry about this later though.

---

btw, I'd like to help write up content to go along with this PR once we have the structured settled. Not sure what the best way to do that is (I PR to your fork probably?) but let me know. Feel free to stub out pages and leave the content writing for later/for me to do.

tomschr:
> Would "Guides" work better? Was looking at the [k8s docs](https://kubernetes.io/docs/home/) and pulled "Guides" from there. "Using" feels like I'm saying half a sentence ("using" what?).

Sure, "Guides" would work too. You are right, "Using" is not a good word.

> [...] A page dedicated to Concourse-specific terms sounds like a good idea, maybe under References? Or Concepts?

I would probably add them into concepts.


> btw, I'd like to help write up content to go along with this PR once we have the structured settled. Not sure what the best way to do that is (I PR to your fork probably?) but let me know. Feel free to stub out pages and leave the content writing for later/for me to do.

Good point! Not sure what's the best approach here. If you think a PR would be good, then we could do that.

Just to summarize the remaining steps what needs to be done:

1. Settle the structure, use "Starts | Guides | Concepts | References" (Tom)
2. Create stub files (Tom)
3. Write content of stub files (Taylor)

Maybe after 1 and 2 is finished you could start with writing the content.

Did I understand it correctly? :slightly_smiling_face:

tomschr:
Ok, I've added some changes to use the new headers ("Starts | Guides | Concepts | References"). So I suppose, point 1 is fixed.

I'm a bit unsure about 2, moving the files into the right folder. The problem is that it might not only to move files from A to B, but also to change the content. Perhaps that could be combined with 3, creating stub files.

Before I continue with this task, I'd like to sync with you and what do you think.

taylorsilva:
> Just to summarize the remaining steps what needs to be done:
>
> 1. Settle the structure, use "Start | Guides | Concepts | References" (Tom)
> 2. Move the files into the right folder (Tom/Taylor?)
> 3. Create stub files (Tom)
> 4. Write content of stub files (Taylor)

Yup, this all sounds good to me 👍
I'm going to move some files around right now and PR to your branch. I'll link back to this PR.

tomschr:
> Seems like we can remove that section?

Maybe, maybe not.

Let's first recapture what a _concept_ really is. According to [Wikipedia](https://en.wikipedia.org/wiki/Darwin_Information_Typing_Architecture#Information_typing):

> Concept information is more objective, containing definitions, rules, and guidelines.

Basically, it answers "Why?" and "What is...?" questions, rather than "How?" for tasks. 

To distinguish them, I've added this useful table from [Kai Weber](https://kaiweber.wordpress.com/2012/03/05/concept-or-reference-whats-the-difference/):

Concept topics | Reference topics
-- | --
Are abstract ideas | Are specific settings
Explain meaning or benefit | Give facts without much explanation
Can stay when specifications change | Change with specifications
Support understanding of tasks | Support correct execution of tasks
Are read for background information | Are read for detailed information
“Why brushing your teeth is important” | “Stages of tooth decay”

For example, we could start with these pages as concepts:

* [Ecosystem](https://concourse-ci.org/ecosystem.html)
* [Intro to YAML](https://concourse-ci.org/config-basics.html#intro-to-yaml)
* [Internals](https://concourse-ci.org/internals.html)
* [Common Pipeline Practices](https://concourse-ci.org/common-pipeline-practices.html)

Perhaps we need to propagate some into a higher level or even we need to split them. Not sure yet as I don't know all of the docs yet.

If we proceed this way, we will certainly come to a point where moving won't get us any further. We need to get our hands dirty and touch the content. Not sure yet how to approach this.

tomschr:
Hmn, another idea... :thinking: ... what about "Learn" or "Learning" instead of "Guides"?
Learn is a verb and more action-oriented whereas "Guides" is a noun and more descriptive.

taylorsilva:
That concept vs reference table is helpful.

Guess most of the work is splitting content between those two sections now. Might be hard to do though as I think concept and reference stuff may be mixed together on various pages right now. Will have to slowly start going through it.

Those pages you listed as concepts make sense to me now and are probably a decent start. I'll start going through more pages.

tomschr:
In the previous commit 7b9742ce, I've added some template. This is how I envision the overall structure of a task, concept, or reference. Some are probably a bit more rigid than others. The overall goal here is consistency. :wink: 

It may or may not turn out to be useful. Perhaps we can refine the templates, if it's useful.

taylorsilva:
this was fun but don't think it's happening. Not enough time for me to review this. Can't believe it's been over a year!

PR #499, opened 2022-05-17
Update doc for https://github.com/concourse/concourse/pull/8184

evanchaoli:
When global resource is enabled, a check container may run on any worker.

### DIFF
--- a/lit/docs/operation/global-resources.lit
+++ b/lit/docs/operation/global-resources.lit
@@ -52,14 +52,15 @@ instead associated to an anonymous 'resource config' i.e. its
       which is workers belonging to a team and workers with tags.
 
       If a resource has \reference{schema.resource.tags} configured, \italic{and} the
-      resource's check interval ends up acquiring the checking lock, a new
-      container will be created on a worker matching the appropriate tags, even
-      if a check container already exists for the same resource config
-      elsewhere.
+      resource's check interval ends up acquiring the checking lock, if a check
+      container already exists for the same resource config elsewhere, it will reuse
+      the container, otherwise a container will be created on a worker matching
+      the appropriate tags.
 
       Similarly, if a team has its own workers, and their check interval ended
-      up acquiring the lock, a new container will be created on the team's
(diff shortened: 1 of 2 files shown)


lit/docs/operation/global-resources.lit
```diff
@@ -52,14 +52,15 @@ instead associated to an anonymous 'resource config' i.e. its
...
+      resource's check interval ends up acquiring the checking lock, if a check
+      container already exists for the same resource config elsewhere, it will reuse
+      the container, otherwise a container will be created on a worker matching
+      the appropriate tags.
 
       Similarly, if a team has its own workers, and their check interval ended
-      up acquiring the lock, a new container will be created on the team's
-      workers, rather than re-using a container from the shared worker pool.
+      up acquiring the lock, it will try to re-use a container from the shared
+      worker pool, rather than creating a new container will be created on the
```
xtremerui, line 62:
```suggestion
      worker pool, rather than creating a new container on the
```

lit/docs/operation/global-resources.lit
```diff
@@ -52,14 +52,15 @@ instead associated to an anonymous 'resource config' i.e. its
...
-      elsewhere.
+      resource's check interval ends up acquiring the checking lock, if a check
+      container already exists for the same resource config elsewhere, it will reuse
+      the container, otherwise a container will be created on a worker matching
+      the appropriate tags.
 
       Similarly, if a team has its own workers, and their check interval ended
-      up acquiring the lock, a new container will be created on the team's
-      workers, rather than re-using a container from the shared worker pool.
+      up acquiring the lock, it will try to re-use a container from the shared
```
xtremerui, line 61:
```suggestion
      up acquiring the lock, it will try to re-use a container with the same resource config from the shared
```

xtremerui:
Thank you for the PR. Could you also remove the `experimental` from the Gloabal Resources page title and add a reference to the team worker complication section on the page from https://concourse-ci.org/concourse-worker.html#team-workers where it says `all workloads` ?

PR #507, opened 2022-10-06
Fix .yaml examples in pipeline guides

michaellihs:
lit seems to ignore the inlined `{repository: busybox}` resources settings. So moved them to a new line and removed the `{` and `}`


taylorsilva (approved):
Thank you! 👏

PR #510, opened 2022-12-17
Updating the runtime worker to *houdini* instead of *containerd*

medChiboub:
The *containerd* runtime has an iptables exception when the image is launched using Docker.  
For more information, please see https://github.com/concourse/concourse/issues/8270 Thank you.


taylorsilva:
This only applies to those running on ARM machines, like M1/M2 macbooks. I'd rather the add a comment to the yaml to tell the user to change the runtime to `houdini` if they're running on an ARM-based machine.

taylorsilva (changes requested):
change to a comment instead for ARM users

medChiboub:
Hello @taylorsilva, 

I just updated the file as you described above. 

Thank you.

PR #538, opened 2024-07-27
Highlite that CONCOURSE_TSA_AUTHORIZED_KEYS is a public key

vladyslav2:
When I read installation guide for a couple times I was sure I need to generate 3 keys and those envs should contain private keys (hey they all without .pub so it must be private keys)
```
  - CONCOURSE_TSA_HOST_KEY
  - CONCOURSE_TSA_AUTHORIZED_KEYS
  - CONCOURSE_TSA_WORKER_PRIVATE_KEY
  ```

ended up reading golang code, creating small script with GO SSH and a similar configuration and ended with understanding that ssh keys does not match.

Ideally, I would suggest to rename those envs but I believe it will might be a breaking changes for old users. So at least, I hope, this will help to improve documentation

keywords for google:
 - concourse ci PTY allocation request failed on channel 0
 - concourse ci shell request failed on channel 0
 - concourse ci tsa.connection.handshake-failed


taylorsilva (commented):
Should update the generate keys page to use the same filename.

taylorsilva (approved):

PR #544, opened 2025-01-16
Note that --cf-org users still need to be part of a Space

taylorsilva:
Resolves https://github.com/concourse/concourse/issues/9052


Kump3r:
Thanks Taylor!

PR #559, opened 2025-05-01
document updated identifier regex

taylorsilva:
This PR can be merged after https://github.com/concourse/concourse/pull/9119 is part of a release.


taylorsilva:
I'm already blocking updating docs because of https://github.com/concourse/docs/pull/563 so might as well merge this in now.

PR #564, opened 2025-07-03
Update observation.lit

jcarrete5:
Added documentation for the title query parameter for the badge web API


taylorsilva (approved):
Thanks for adding that! Inadvertently found out the tag doesn't expand to fit the custom title 😬 Any title longer than 5-6 characters overflows.

Thanks again for the PR!

jcarrete5:
I also noticed that the title doesn't always fit 🙂. How long does it usually take for the docs site to deploy with new changes? @taylorsilva 

I might be able to find some time to submit a PR fixing the expanding title issue as well although it's a pretty minor issue.

PR #569, opened 2025-09-25
docs: Add syslog drainer documentation

### LINKED ISSUE #488: Document syslog format
From a discord discussion, someone was having trouble getting syslog working. Ended up being the expected format on the collectors end.

Apparently the Concourse syslog drainer exports logs in RFC3164. We should document this.

analytically:
Document configuration and usage of the syslog drainer for forwarding build logs to external syslog servers via TCP, UDP, or TLS transport.

Fixes #488

### DIFF
--- a/lit/docs/operation/syslog.lit
+++ b/lit/docs/operation/syslog.lit
@@ -1,0 +1,367 @@
+\title{Syslog Integration}{syslog}
+
+\use-plugin{concourse-docs}
+
+Concourse supports forwarding build event logs ('draining') to external syslog servers, enabling centralized log management, compliance auditing, and integration with enterprise logging infrastructure. When configured, Concourse will stream build events to your syslog server in RFC 5424 format.
+
+\section{
+  \title{Enabling Syslog Draining}
+
+  To enable syslog draining, configure at minimum the syslog address and transport on the \reference{web-node}:
+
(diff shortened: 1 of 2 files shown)


lit/docs/operation/syslog.lit
```diff
@@ -0,0 +1,367 @@
...
+    \title{Verifying Configuration}
+
+    Check that the web node started successfully with syslog draining:
+
+    \codeblock{bash}{{{
+    # Check web node logs for syslog drainer initialization
+    grep -i "syslog\|drainer" /var/log/concourse/web.log
+
+    # Verify the component is configured (look for ComponentSyslogDrainer)
+    curl http://web-node:8080/api/v1/info
```
taylorsilva: LLM hallucination? Internal components are not listed in the info endpoint. See:

- https://github.com/concourse/concourse/blob/750ce8b641a756fb26caa1571d91f8e0955f2ab8/atc/api/infoserver/info.go#L14-L19
- https://github.com/concourse/concourse/blob/master/atc/feature_flags.go

taylorsilva (changes requested):
Okay I think I've validated that most of the info here is correct. I mostly cross-referenced by looking at the syslog code.

For transparency, can you list the LLM(s) used to assist you with writing this PR? TY!

PR #586, opened 2026-02-21
docx: reformat metrics to fit overall site theme and remove .old

kcbimonte:


kcbimonte:
Removes the `.old` folder since we're seeing success with MkDocs. 

Uses the [hooks functionality](https://www.mkdocs.org/user-guide/configuration/#hooks) to parse the Prometheus metrics like the old booklit site.

@taylorsilva I think this is good to merge whenever now

image

taylorsilva:
That metrics page is so much nicer!

taylorsilva (approved):

PR #604, opened 2026-05-05
resource-types: Update irccat

kallisti5:
* The docker image hasn't been used for some time
* Update to ghcr.io (and update image base)
* Docs updated upstream for verification of new location


taylorsilva (approved):
TY!

PR #606, opened 2026-05-12
Add page for available icons, do some JS cleanup

jghiloni:
Adds a page *Available Resource Icons* to the root navigation next to *Resource Types*. It behaves similarly in its search functionality with a few notable differences:

* Because there are over 21000 icons now available, it uses the Mkdocs Material Grid layout instead of a table
* You can click a grid card to copy the name of the icon to your clipboard

In addition to adding that functionality, I made some improvements to the JS developer experience in the page:

1. `list.js` is now an NPM dependency, rather than having to keep the minified copy / source map in overrides. I copied and improved the Typescript type definitions from `@types/list.js` into `src/types` to make it usable in Typescript.
2. `list.js` also was causing an issue in the rollup build, due to its structure. By adding the rollup `commonjs` plugin, it ensures that all modules are at least CommonJS modules before building, allowing it to complete with no errors.
3. This allowed me to move the script at the end of `resource-types-list.md` to the root script bundle with a guard for the specific DOM elements existing.
4. I was having significant issues debugging JS with the source maps created, because they would not be available when running locally via `npm run start`. I created `debug` equivalent scripts that `npm run start` calls that does not generate the source map OR minify the code, allowing the debugging to be much smoother.
5. I updated a couple of package dependencies (specifically terser) that were reporting High security issues
6. I did not include the generated list of icons because it's large but `tools/icon-list/generate-icon-list.mjs` runs in subsecond times on my macbook air.
7. While I wanted to peg the `simple-icons` dependency to the version we use in concourse, `mkdocs-material` uses an older version and the new version has a non-trivial number of new icons that don't render. I pegged to the version in `mkdocs-material` and added a footnote.


docs/docs/resources/available-resource-icons.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+# Available Resource Icons
+
+This is a list[^1] of icons and their names that you can use in your pipelines' `resource`
+definitions. Hover over an icon to get the name of the icon or click it to copy the name to 
+your clipboard. You can also search using the text field below, with or without the `si/`
+prefix.
+
+<div id="icon-list" markdown>
```
jghiloni: both this div and the div on line 17 have the `markdown` attribute intentionally -- if you want to render markdown in a nested element, it's gotta be annotated from the root down to the actual container element

jghiloni:
If accepted, this will also close #596

PR #613, opened 2026-07-10
add: Health endpoint documentation

Kump3r:
Follow-up from https://github.com/concourse/concourse/pull/9588

### DIFF
--- a/docs/docs/operation/health.md
+++ b/docs/docs/operation/health.md
@@ -1,0 +1,287 @@
+---
+title: Health Endpoint
+---
+
+## `GET /api/v1/health`
+
+The health endpoint is an HTTP endpoint exposed by the [`web` node](../install/running-web.md) that reports whether
+the Concourse deployment is in a healthy state. Operators and monitoring tools can poll it to get a fast, structured
+answer about the state of the system — without needing to log in or use `fly`.
+
+The endpoint requires no authentication and returns a JSON body along with an HTTP status code that reflects the
+current health state.
+
+## Health states
+
+The endpoint reports one of three states:
+
+**Healthy** — everything is working as expected. The number of connected workers meets the configured minimum, and
+all internal components are keeping up with their scheduled work.
+
+**Degraded** — the system is running but something needs attention. Either the number of connected workers has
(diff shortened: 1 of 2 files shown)


taylorsilva (commented):
LGTM - Added a commit changing a few minor things.

Will merge once https://github.com/concourse/concourse/pull/9640 is merged and we're sure the docs here conform with the output of the `fly health` command.

Kump3r:
> LGTM - Added a commit changing a few minor things.
> 
> Will merge once [concourse/concourse#9640](https://github.com/concourse/concourse/pull/9640) is merged and we're sure the docs here conform with the output of the `fly health` command.

Thanks for the quick reviews!

docs/docs/operation/health.md
```diff
@@ -0,0 +1,293 @@
...
+    { "name": "collector_volumes", "healthy": true, "paused": false, "last_ran": "2026-06-08T10:05:00Z", "stale": false }
+  ]
+}
+```
+
+### Degraded
+
+```
+HTTP 333 Degraded
+```
```
taylorsilva, line 90: Was reviewing https://github.com/concourse/concourse/pull/9640 and noticed the client code didn't handle an HTTP 333 response. I then realized I don't recall this status code being a thing in the original PR. I looked at the health endpoint handler and I'm not seeing where we return `HTTP 333`: https://github.com/concourse/concourse/blob/master/atc/api/healthserver/health.go

I don't think HTTP status codes were discussed in the RFC as well, which is fine. This section here should be updated to reflect the code OR the code should be updated to reflect the docs here. I'm a bit worried about returning a non-standard HTTP status code, so updating the docs to reflect the fact that we actually return HTTP 200 here is probably what we should do.

taylorsilva (changes requested):

docs/docs/operation/health.md
Kump3r: Yes, I just forgot to fix this I guess. I had some stuff noted down during testing different approaches and reading trough articles and pushed an older version of the docs, not the final one. Silly me, good catch. There were also other things that were not from the latest `health` service merged. I will commit the most recent version. Sorry for the noise.

taylorsilva (approved):
LGTM - We'll merge when the next release is out.

PR #616, opened 2026-07-31
docx: update credentials manager docs with new features

kcbimonte:


docs/docs/operation/creds/conjur.md
```diff
@@ -123,13 +63,61 @@ the [official documentation](https://docs.conjur.org/Latest/en/Content/Operation
...
+
+When secrets are to be looked up, these are evaluated where `{{.Team}}` expands to the current team, `{{.Pipeline}}` to
+the current pipeline (if any), and `{{.Secret}}` to the name of the secret. So if the settings are:
+
+```properties
+CONCOURSE_CONJUR_TEAM_SECRET_TEMPLATE="{{.Team}}/concourse/{{.Secret}}
+CONCOURSE_CONJUR_PIPELINE_SECRET_TEMPLATE=/{{.Team}}/concourse/{{.Pipeline}}/{{.Secret}}
+CONCOURSE_CONJUR_SECRET_TEMPLATE=conjur/{{.Secret}}
+CONCOURSE_CONJUR_SHARED_PATH=common
+```
```
kcbimonte, line 102: Is Conjur Secret Template intended to be the Shared Secret Path? If so, is concourse/concourse#9651 needed or should we revert?

taylorsilva (approved):
Thanks Kevin! This is a really nice clean up of all these pages too. They feel more concise and easier to read.

taylorsilva:
Appreciate you adding it to the milestone as well.

docs/docs/operation/creds/conjur.md
kcbimonte: @taylorsilva, what do you think of this part? Don't necessarily want two config properties that do basically the same thing

taylorsilva: OH interesting! The main difference between the two seems to be `CONCOURSE_CONJUR_SECRET_TEMPLATE` doesn't start with a forward slash `/`. I don't know what the means within Conjur itself. Maybe it means nothing and this was just an implementation quirk? It definitely reads like a "share path" config.

Tried looking at the Conjur docs but their site is completely busted right now. Whole project looks kinda dead actually. Ugh feeling like a good example of code I wish we didn't have to maintain.

Looking at git history where this was introduced: https://github.com/concourse/concourse/commit/bd7109e5417a359a050d6cec81faf5b43bc9b5aa

So much of this code is unchanged from its initial implementation. I think the intention is that it can be used for shared secrets, the naming convention is just off from all the other managers.

I'm going to open a PR reverting the new flag that was added. I'll ping the person that initially added `CONCOURSE_CONJUR_SECRET_TEMPLATE` and see if they can explain how it's meant to work.

Thanks for calling this out Kevin. Really good catch!

PR #619, opened 2026-08-18
Link to new blog URL

jandd:
Link to https://concourse-ci.org/blog/ instead of the old https://blog.concourse-ci.org/


taylorsilva (approved):
Thanks for fixing that!

